PR #2, opened 2016-02-06
fix clippy warnings

llogiq:
This includes some minor readability fixes and removes a few needless `vec!`s and closures.


sseemayer:
Awesome, already learned a bunch from this! Thanks!

PR #4, opened 2016-04-01
Try quick-xml instead of xml-rs

tafia:
I've tried changing xml-rs with [quick-xml](https://github.com/tafia/quick-xml) to see how it performed in your *simple* scenario

```
before
test bench_sample ... bench:   6,658,267 ns/iter (+/- 509,266)

after
test bench_sample ... bench:   5,008,046 ns/iter (+/- 108,389)
```

quick-xml is a library I'm building which if definitely not as bullet proof as xml-rs. But for many scenario it is really faster.

Warning: there are many `unwrap`s  in the code (you were already unwrapping event).


sseemayer:
Welcome and thanks for the PR! Am I reading your benchmark correctly in that using quick-xml will be `1.7*10^6` nanoseconds aka 1.7 milliseconds faster per iteration?

I'm currently expecting the main runtime in opening a KeePass database to come from [deriving the decryption key](https://github.com/sseemayer/keepass-rs/blob/master/src/decrypt.rs#L87-L112) (i.e. doing many iterations of AES256) and not from parsing the decrypted XML since most users will have much less than 1000 entries in their database - though I don't have any data to back this up :)

I admit that the demo database included with keepass-rs is probably smaller than a typical user's real database, but do you see a realistic use case where the XML parsing will be a significant portion of the run time, thus making a faster parser necessary?

tafia:
I was actually surprised by the benchs (was expecting around 0.2 - 0.3ms improvement max).

I don't know keepass enough to be certain be yes, I agree it probably doesn't matter that much. I took this as an exercise to see how difficult it was to switch from one library to the other.

sseemayer:
Since the performance difference between the two parsers is not that important for this project, I prefer to stick with the more widely used XML parser library. 

Thanks for pointing out `quick-xml` to me though, I'll keep it in mind for future projects - and I agree that porting between the two libraries is very easy indeed :+1:

PR #13, opened 2019-07-01
update IncorrectKey Display msg

jxs:
Hi there,
at first usage I could not assert that `IncorrectKey` `Display` was meaning the password instead of the key file, what do you think of specifying that it's the password?
thanks


sseemayer:
Sorry for the late reply -- A lot of the innards of the library have changed, including the error handling. I hope the [new error handling](https://github.com/sseemayer/keepass-rs/blob/master/src/result.rs) is useful.

PR #17, opened 2020-01-21
Update flate2 dependency and fix breaking change

Narigo:
I've been playing around with WebAssembly lately and the [flate2](https://github.com/alexcrichton/flate2-rs) dependency in its currently used version `0.2` seems to be problematic for wasm-pack to compile this crate. The `flate2` code has been [fixed](https://github.com/alexcrichton/flate2-rs/issues/161) and a `1.0` version of `flate2` exists.

I went ahead to update it and see if anything breaks. There is one occurrence in the code where they changed the return type of a function and got rid of a `Result<X>`, making a small change necessary.

I guess the `Cargo.lock` changes come from the dependency updates in `flate2`.


sseemayer:
Sorry for the late response and thanks for the contribution!

PR #22, opened 2020-05-10
Bump dependencies

linkmauve:
Surprisingly, none of them required any code change.


kraag-gorim:
I think the `^` notation should be kept.

linkmauve:
No, see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

More specifically, “The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.”

kraag-gorim:
Yeah I looked it up. I still don't get, why you removed it.

linkmauve:
Mostly just for consistency with the other dependencies, which already didn’t have it.

sseemayer:
Thanks a lot for this!

PR #23, opened 2020-05-29
KeePass 1.x KDB support

laomaiweng:
This PR implements support for KDB (KeePass 1.x) databases. The crypto is similar to that of KeePass 2.x, but the header and actual DB format differ.

Header parsing and crypto are based on: https://gist.github.com/lgg/e6ccc6e212d18dd2ecd8a8c116fb1e45
DB parsing is based on the KeePass 1.10 sources.

I tested the feature with a password-protected DB (testcase) and a keyfile-protected one.
I also tried to add support for binary data fields by adding a variant to `Value` (see 99e54e2), but I'm not sure that's the best way to go about it.
I should mention that I'm a Rust noob, so feel free to suggest improvements!


sseemayer:
Woo-hoo, thanks a lot for this! That's an awesome feature addition!

PR #26, opened 2020-08-28
Fix Clippy lints

strangelittlemonkey:
* Remove unneeded formats, prefer .to_string()
* Remove redundant field name in struct
* Satisfied must_use by discarding other half of vec explicitly rather
  than implicitly


OliverEvans96:
This seems like it should be a pretty easy one to merge. Any opposition from maintainers?

jgrussell:
I started to make some of these changes in my local copy before seeing this PR.

Any thoughts on when this PR might be merged here, @sseemayer?

sseemayer:
Sorry to have slept on this - merged.

PR #27, opened 2021-02-26
Add github workflows

scottmmjackson:
This adds github workflows. It doesn't delete the travis.yml. It can be amended to do so. Submitted for author feedback.


sseemayer:
Sorry for the late reply to this. I will merge right now to see how well github works compared to Travis and compare both for a while.

Issue #24, opened 2020-07-04
Group or entry with same title does not get into iterable


Narigo:
When having the same title twice in a group, only one will show up in the groups `iter()`. The other group or entry won't show up at all. It does not seem to be accessible through other means as well. Is there something I'm missing?

**Additional context**
I was trying to create a reproducible test for #16 - looks like I can't create an empty title (only `Some("")`). Anyways, trying to come up with such a test unearthed this issue.

OliverEvans96:
I suspect it's because `Group` stores its entries in a `HashMap<String, Entry>` with titles as keys, and the first entry is overwritten when the second one is read.

https://github.com/sseemayer/keepass-rs/blob/32dd3c8b44b47b5fc19d614bbbb17ba6592e348e/src/db.rs#L274-L284

A few options that come to mind:
1. `entries: HashMap<String, Vector<Entry>>`
2. `entries: HashMap<String, Entry>`, but with a unique key (I'm not familiar with the keepass format - is there a unique key id?)
3. `entries: Vector<Entry>` instead of a `HashMap<String, Entry>`, and do lookups by looping over the whole entry. Seems similar to 2., but less efficient.

Anyone else have thoughs on this?
Oliver

OliverEvans96:
Actually, the best of both worlds might be to store the values in `entries: Vec<Entry>`, but also have `title_map: HashMap<String, Vec<usize>>` that stores the indices of the elements of `entries` by title. We could have efficient iteration _and_ lookup.  Since `title_map` can be created during parsing at the same time that `entries` is being filled, it shouldn't introduce much execution time overhead, and the memory usage should also be quite small. 

I put together a little simplified demo:

```rust
use std::collections::HashMap;

struct XMLNode(String, String);

#[derive(Debug)]
struct Entry {
    title: String,
    password: String,
}

impl From<XMLNode> for Entry {
    fn from(n: XMLNode) -> Self {
        Entry {
            title: n.0,
            password: n.1,
        }
    }
}

#[derive(Debug)]
struct Group {
    entries: Vec<Entry>,
    title_map: HashMap<String, Vec<usize>>,
}

impl Group {
    pub fn new() -> Self {
        Self {
            entries: Vec::new(),
            title_map: HashMap::new(),
        }
    }

    pub fn get<'a>(&'a self, key: &str) -> Option<Vec<&'a Entry>> {
        match self.title_map.get(key) {
[...]

PR #30, opened 2021-04-03
Add support for native group order

### LINKED ISSUE #28: Preserve order (BTreeMap?)
The order is lost on groups when iterating.

**Disclaimer**: I've not looked at the internals of this crate nor the keepass format.

Assuming this crate reads everything in order from the .kdbx file, a `BTreeMap` would retain said order. If not, perhaps there is an order index somewhere on each group which could help?

leaty:
Fixes #28 

I made use of the crate [indexmap](https://github.com/bluss/indexmap) to retain the original insert order of groups found in the kdbx file. This means that iterating over `child_groups` is sorted just like the original file (and is identical to how it looks in keepass frontends). I've not tested this in anything but a 4.0 kdbx file however, but as long as they're read and inserted in the same order, other versions should work identically. 

Note: This does not automatically make `NodeIter` sort perfectly, but doing a `self.queue.remove` of the first element instead of `self.queue.pop` works. I did not include it though, as I'm not sure if something like that is wanted - I don't personally use it as I need a more true recurse to retain parent/child relationships.

Note 2: I chose to exclude entries from this change, as I do not know if entries are supposed to have a native order. In the frontend that I use - `keepassxc`, you're always forced to sort entries by something (there is no manual index order).

Thanks a lot for this crate, it helped me to write a persistent keepass backend (shameless plug: [rkeep](https://github.com/leaty/rkeep)) for [rofi](https://github.com/davatorium/rofi).

Let me know if something should be changed!


sseemayer:
Thanks a lot for the PR, and for the explanation! I read the code first and had exactly the two questions that you already wrote notes for 😆 

If you don't mind, could you also make the proposed modification on the `NodeIter`? For people using the library, I think it would be less confusing if both ways of accessing yield the expected order of groups.

leaty:
No probs! I'll do that later today.

I had a thought about entries, do you think it'd be worth storing them in a BTreeMap? That way the entries are at the very least sorted consistently (alphabetically), so that each run has perfectly consistent iterations. You could still sort them as you please, but at least it''s a sane default.

sseemayer:
BTreeMap sounds good :) I expect most people will want alphabetic sorting regardless.

FYI, I also still have #24 on my to-do list, so I am planning to touch your changes again right after your MR is merged to fix this issue, then release a new crate version.

leaty:
I'll make a new PR with those added changes.

PR #31, opened 2021-04-06
Add support for native order

### LINKED ISSUE #28: Preserve order (BTreeMap?)

leaty:
Fixes #28
Continuation of #30 

I implemented both additional "notes" according to above PR. Fun fact, I found out that you actually *can* "unsort" entries, whereby they automatically get sorted by last added (created). This is done by clicking the currently sorted column in e.g. `keepassxc` until the arrow disappears. Entries are also stored like this in the kdbx file, thus I chose `IndexMap` here again, as it *technically* makes the most sense.

If you think `BTreeMap` (alphabetical) would be better I'll just change it (it's literally 2 lines), but for correctness sake `IndexMap` does seem right here!


sseemayer:
Merged. Thanks a lot for the PR! 

I think it would have also worked to push additional changes to the already-existing PR, but this works just as well :)

leaty:
I just didn't like the branch name as it didn't completely fit anymore :)

Issue #33, opened 2021-04-07
Not Pulling Some Data from KDBX4 File Versions


jgrussell:
`Entry::AutoType.sequence` is one example of data that is not being populated from KDBX4 files but is working fine with KDBX3 files.

I actually stumbled onto this issue attempting to pull ExpiryTime from the database for #29.  My changes in `xml_parse.rs` seem to be working great with KDBX3 files but showing gibberish for this XML value in KDBX4.

Sadly, I do not yet have the rust skills address this myself.  But, I am putting together a draft PR updating `entry_tests.rs` and `test_db_kdbx4_with_password_aes.kdbx` to illustrate the Entry::AutoType.sequence behavior.

sseemayer:
Thanks for filing the issue! Probably, there is some difference in the XML formats betwen KDBX3 and KDBX4 that I was not yet aware of.

sseemayer:
As a heads-up, I am in a state of being constantly very busy with my day job, so I might not be able to do much troubleshooting on this right now. It might take me some time to fix this.

jgrussell:
I think I have found my issue with pulling `ExpiryTime` from the database for #29 at the very bottom of the official [KDBX 4 documentation page](https://keepass.info/help/kb/kdbx_4.html):
> In encrypted KDBX 4 files (not in plaintext XML exports), all times are now stored as Base64 string of the Int64 number of seconds elapsed since 0001-01-01 00:00 UTC. This improves the performance and reduces the file size.

Note:  The file formats look the same when saved/exported as unencrypted XML!

@sseemayer,
I would love to take you up on the mentoring offer as your availability permits.  I think I am getting close to having something presentable via a Draft PR for addressing #29 and will definitely appreciate any corrections, commentary, suggestions, etc.  I have tried to follow your coding style for a few reasons:

- This is my first Rust coding (also my first GitHub experience) of any significance; and, assuming we complete it, my first real Open Source project contribution.
- I am a firm believer that coding styles should at least be consistent within modules, ideally within projects.
- Hopefully, this will make the code and any later maintenance easier.

Maybe I can track down the KDBX4 issue with `Entry::AutoType.sequence` after that.

Thank you again!

sseemayer:
Sounds great, let's do it! 💪 

regarding coding style in Rust, there is [rustfmt](https://github.com/rust-lang/rustfmt). Just

1. run rustfmt on the codebase before you do any work
2. if there are changes, commit them separately
[...]

PR #34, opened 2021-04-07
Updating Tests for KDBX4 File Issue

jgrussell:
If you want to write your tests differently, I will remove this PR.  If you do want this change, I will mark it as ready to review and merge.

This is the Draft PR promised when #33 was created.

Please let me know if you think I can be of any additional assistance with this.


sseemayer:
Thanks a lot for the PR and filing the related issue! This looks good to me to add a test like this.

PR #36, opened 2021-04-14
Add Expire Timestamp and Related Information

jgrussell:
This PR is definitely not ready to be merged; I just wanted to let you see where I was going to see if you have any issues, recommendations, etc. with my approach...and to ask for some specific help, hopefully just a second set of eyes that will quickly notice something amiss.

I already have things working well enough to use for my personal pet Rust project since my main KeePass db is still 3.1 format which is working as expected, illustrated in `tests/entry_tests.rs`; but, I have been stuck trying to figure out how to convert kdbx4 time nodes into something useful.  My new `kdbx4_entry fn` in `tests/entry_tests.rs` illustrates where I am stuck:

`e.times.get_time("ExpiryTime")` contains the raw XML character string for this node pulled from `parse_xml_block`.  (This is working fine for getting human readable results from kdbx3 files but not so with kdbx4 files.)

I tried converting this kdbx4 field from the Base64 encoded entry as follows; this code is currently in my new `get_time()` in `db.rs`:
1) Get vec of u8 digits representing the Base64 digits using `base64::decode` which is already being used elsewhere in the module.
2) Multiply out the vec digits to get a u64 (normal base 10 integer) representation of this value.  

I had assumed this would give me the number of seconds elapsed since 0001-01-01 00:00 UTC as documented at the bottom of the [KDBX 4 release page](https://keepass.info/help/kb/kdbx_4.html).  But, the number of seconds I am calculating is a couple of orders of magnitude smaller than the [EpochConverter page](https://www.epochconverter.com/seconds-days-since-y0) says I should be seeing.

I did glance through the [KeePassXC](https://github.com/keepassxreboot/keepassxc) code to see how they are handling this; but, at first glance, it looks to me like they may just be handing off the conversion work to a QT library.  However, it is very possible I lost may way in that code; it has probably been 25 years since I wrote any C++.

If nothing jumps out at you, please don't feel like you have to spend a lot of time helping me with this.  I am actually hoping you will quickly see something that I have been staring at and missing, or completely misunderstanding, for the last few days.  If not, I can keep digging on my own.

Note:  My approach to handling ExpiryTime is a bit different than I had originally planned; but, I think this approach lays the groundwork for easily and consistently handling all time time fields from the KeePass XML.  If you are good with this approach (once it is actually working), I will do the same for all custom data fields; currently, I just have enough code in place to pull out the db_report_exclude flag that I need.


jgrussell:
@sseemayer or anyone willing to help me out:

I am not sure if I am missing something fundamental or have just made an obvious mistake that I am just not seeing.

I have run the following code for all possible values of c.  The combinations that return Ok() produce one of the following values of d:
291529854
861071171194951
141427661638534

The [EpochConverter site](https://www.epochconverter.com/seconds-days-since-y0#s1) indicates I should be seeing something on the order of 6xxxxxxxxxx for the number of seconds after 0001-01-01, "/s4D2A4AAAA=" representing April 10, 2021 in the test_db_kdbx4_with_password_aes.kdbx file's XML.

```rust
use std::convert::TryFrom;

fn f(c: base64::Config) {
    match ::base64::decode_config("/s4D2A4AAAA=".as_bytes().to_vec(), c) {
        Ok(v) => {
            let mut d: u64 = 0;
            let b: u64 = 64;
            for (i, e) in v.iter().enumerate() {
                d += u64::pow(b, u32::try_from(i).unwrap())
                    * u64::try_from(*e).unwrap();
            }
            println!("d = {} for v = {:?} from c = {:?}", d, v, c)
        }
        z => println!("c = {:?} fails with {:?}", c, z) 
    }
}
```

jgrussell:
I think I have finally figured out how to get a u64 from the Base64 encoded integer; so, I am pulling this PR while I clean up my code and test.

Issue #29, opened 2021-04-03
Add "Expires" struct to both Group and Entry


jgrussell:
I am planning to use this crate for a small personal project not only to familiarize myself with Rust but also to use as a [hopefully] short term solution to this KeePassXC feature request that has been around in one form or another for several years.
https://github.com/keepassxreboot/keepassxc/issues/4624

In my limited testing so far, I have found this crate very easy to use and robust.  However, it does not provide access to the Group or Entry expiration information which I will need for my project.  If there is no interest in merging something like that into this crate, I will only do enough work to fulfill my own needs.  However, I think this might be useful to others; so, I am more than willing to fully build this feature out for inclusion here.

My initial plan is to loosely model a new Expires struct and associated population on the existing AutoType struct.  

Additionally, I will probably also want to add an Exclude from database reports indicator on Entry.  For this, I am thinking a new bool in Entry would suffice.

I welcome (and will appreciate) any feedback on both the likely inclusion of these features into this crate as well as my planned approach.

Thank you.

sseemayer:
This sounds good to me, and I would be willing to merge something like this.

How is the Expires field encoded in the KDBX? Is it a XML element similar to how AutoType works?

jgrussell:
Expires looks to be encoded similarly to a very basic AutoType, without AutoType's complication of allowing multiple Window Associations, each window optionally allowing for its own auto-type sequence.

If my cursory analysis holds, my new Expires should be much simpler to implement than your AutoType.

I haven't yet looked through all of your existing test .kdbx files; but, if none of them have groups and entries with expiration dates set, do you prefer that I update existing .kdbx files (assuming I can do so without breaking your existing tests) or create new .kdbx files to support my new integration tests of this new functionality?

Finally, thank you so much for both providing this crate and for the quick responses to this issue.

sseemayer:
Sounds good! I don't see an issue with modifying the existing test .kdbx files since they don't check the expiration logic in any way.

Thanks to you for the great idea for another feature :)

jgrussell:
I have created #36 showing where I am going with this and asking for some specific help with converting a Base64 encoded elapsed seconds value to a u64 value.

Thank you.

jgrussell:
#38 adds this functionality.

PR #38, opened 2021-04-22
Add times and expires to Entry and Group

jgrussell:
This adds the `ExpiryTime` entry from the XML; and, I think it sets the stage for quick and easy additions of other XML timestamp fields as needed/requested.

It also adds and the `Expires` flag from the XML to both `Group` and `Entry` since `ExpiryTime` is almost meaningless without this flag.

Please let me know if you want me to make any changes, anything from code tweaks to overall approach; or, feel free to edit my fork directly.

Once this is merged,  #29 will be complete in my mind.

Thank you again!


src/xml_parse.rs
```diff
@@ -9,13 +9,39 @@ use xml::reader::{EventReader, XmlEvent};
...
     Group(Group),
     KeyValue(String, Value),
     AutoType(AutoType),
     AutoTypeAssociation(AutoTypeAssociation),
+    ExpiryTime(String),
+    Expires(bool),
+}
+
+fn parse_xml_timestamp(t: &str) -> Result<chrono::NaiveDateTime> {
+    match chrono::NaiveDateTime::parse_from_str(t, "%Y-%m-%dT%H:%M:%SZ") {
```
sseemayer, line 26: In ISO 8601, [`Z` suffix is for specifying "Zulu" time](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators). Can you use [`chrono::DateTime::parse_from_rfc3339`](https://docs.rs/chrono/0.4.0/chrono/struct.DateTime.html#method.parse_from_rfc3339) instead?

src/db.rs
```diff
@@ -283,6 +284,12 @@ pub struct Group {
 
     /// The list of entries in this group
     pub entries: IndexMap<String, Entry>,
+
+    /// The list of time fields for this group
+    pub times: HashMap<String, chrono::NaiveDateTime>,
```
sseemayer, line 289: Are there other times variables other than ExpiryTime possible? Update: Yes, for creation, last modified, etc.

From a developer usability perspective, I would prefer `expires` to be an `Option<chrono::DateTime>` but I understand that it is mostly the KDBX format that is strange here. I leave it up to you what solution you like better.

src/db.rs
```diff
@@ -321,6 +328,14 @@ impl Group {
...
         }
     }
+
+    /// Get a timestamp field by name
+    pub fn get_time(&self, key: &str) -> Option<&chrono::NaiveDateTime> {
+        match self.times.get(key) {
+            Some(ref t) => Some(t),
+            _ => None,
+        }
+    }
```
sseemayer, line 338: similar to the `get_username` etc. methods, a `get_expiry_time` method would be a nice to have.

src/db.rs
```diff
@@ -321,6 +328,14 @@ impl Group {
             }
         }
     }
+
+    /// Get a timestamp field by name
+    pub fn get_time(&self, key: &str) -> Option<&chrono::NaiveDateTime> {
+        match self.times.get(key) {
```
sseemayer, line 334: Since `get` already should return an `Option<&T>`, you should just be able to write

```rust
pub fn get_time(&self, key: &str) { 
    self.times.get(key)
}
```

This might need some additional `.as_ref()` , `.as_deref()` , etc . if you run into type checking issues.

src/xml_parse.rs
```diff
@@ -9,13 +9,39 @@ use xml::reader::{EventReader, XmlEvent};
...
+        Ok(ndt) => Ok(ndt),
+        // In KDBX4, timestamps are stored as seconds, Base64 encoded, since 0001-01-01 00:00:00
+        // So, if we don't have a valid ISO 8601 string, assume we have found a Base64 encoded int.
+        _ => {
+            let v = base64::decode(t).map_err(|e| Error::from(DatabaseIntegrityError::from(e)))?;
+            // Cast the Vec created by base64::decode into the array expected by i64::from_le_bytes
+            let mut a: [u8; 8] = [0, 0, 0, 0, 0, 0, 0, 0];
+            for (i, u) in v.iter().enumerate() {
+                a[i] = *u;
+            }
```
sseemayer, line 37: `v` has the type `Vec<u8>`. You could use [`a.copy_from_slice(&v[0..8])`](https://stackoverflow.com/a/37669259) with a slice of `v` as an argument to avoid the for loop (not too much of an improvement, I admit)

sseemayer (commented):
Beautiful work on this PR 👍 ! I have some minor comments but this is already on a great way to getting merged.

src/db.rs
jgrussell: I will add this method.

src/db.rs
jgrussell: I will see if I can make this work and either close this or comment further if I need assistance.

src/xml_parse.rs
jgrussell: I will see if I can make this work and either close this or comment further if I need assistance.

I will admit to struggling a bit with converting Vec<u8> into [u8; 8]; and, I just declared success and moved on after finding something that worked.

sseemayer: I've also run into the same bottleneck in the language before. The solution you have come up with is completely fine. Since you wanted some mentoring, I am being a bit nit-picky and just want to point out some opportunities for expressing things slightly more elegantly.

src/xml_parse.rs
jgrussell: I am very happy to have a second set of eyes and sounding board for the approach you see here.

Background on how I arrived at this design:

KeePassXC (and, I believe the original KeePass itself) stores timestamps, at least the ExpiryTime, with the Z timezone indicator.  But, what they are actually storing is the local machine time not converted to Zulu/UTC time.  This is neither ISO 8601 compliant (where the timezone is optional) nor RFC 3339 compliant (where a timezone is required and intended to be accurate).  Sadly, I have seen this approach more times than I would like to admit; so, it did not surprise me.

I tested this with KeePassXC on a machine with the time zone set to CDT (UTC−05:00).  The ExpiryTime was stored with the Z timezone indicator but not actually converted to UTC.  

chrono::DateTime::parse_from_rfc3339 returns a DateTime<FixedOffset>, this would accurately represent what is stored in the XML document text but might mislead anyone using this crate.  

chrono::NaiveDateTime::parse_from_str returns a NaiveDateTime which seemed to me a better representation of these timestamps for users of this crate.  (NaiveDateTime does not have a parse_from_rfc3339 method.)

Please let me know how you want me to proceed with this:
1) Use chrono::DateTime in this crate (and ::parse_from_rfc3339 here) instead of chrono::NaiveDateTime
2) Stick with the current design and code
3) More testing (other KeePass clients, other timestamp fields such as Created and Modified, etc.)
4) Comment reasons for this approach more thoroughly
5) Other (If so, please elaborate.)

sseemayer: Oh wow, that is a very footgun-style design. In this case I fully agree with your decision and would stick with your code as-is. I would only request some additional docs for the appropriate fields in the `Entry` struct [so that they show up in the generated docs](https://docs.rs/keepass/0.4.7/keepass/struct.Entry.html) (i.e. with a `///` comment)

src/db.rs
jgrussell: I coded expires as an Option<> in one of my earlier iterations; however, as you already discovered, there are several timestamps stored in the XML for each Entry and Group (as well as for Database and maybe other XML nodes) which might be useful to others.  Some examples, not an exhaustive list, these may not be the exact XML node names but are close:
CreatedTime
LastAccessedTime
LastModifiedTime

I am only currently populating ExpiryTime for a few reasons:
To minimize size of Database created by keepass::Database::open
To minimize XML parsing time for keepass::Database::open
To minimize my coding and testing effort before you had reviewed this approach

My thought was storing these as a HashMap<> (similar to Entry.fields) sets the stage for quick and easy addition of other XML timestamp fields as needed/requested or if you decide to make the keepass::Database a more exhaustive representation of the XML in the future.

Adding the getter convenience methods (get_expiry_time, etc.) seems like it would provide similar developer usability as using individual attributes.

If you would rather have all timestamps stored unique Option<> attributes rather than in one big HashMap, I can make that change.  Please let me know.

Also, do you want me to add any other timestamps to Entry and/or Group at this time?

jgrussell:
@sseemayer 

Thank you for the compliment and the quick review.

Please take a look at these two conversations and comment further when you have time:
parse_xml_timestamp
pub times: HashMap

I will work on these changes over the weekend and resolve the associated conversations as I progress:
Add get_expiry_time method.
Suggested get_time method changes.
Suggested Vec<u8> into [u8; 8] changes.

src/db.rs
sseemayer: OK, that absolutely makes sense. Let's keep it as it is right now. You don't need to add the other timestamps right now, since it is not something that you need for your own use case. I can see that @Narigo might want to know about Entry modification timestamps for keepass-diff for a future automatic conflict resolution, but let's keep that for a future PR.

sseemayer:
Thanks also to you for the elaborations! I've added some further comments on the parse_xml_timestamp and pub times: HashMap discussions.

src/db.rs
jgrussell: I am part of the way there.  The following code works; but, the compiler cannot infer the return type of the method if I leave off the `-> Option<&chrono::NaiveDateTime>`. 

Are you thinking I should be able to eliminate this explicit return type declaration as well?

```
/// Get a timestamp field by name
pub fn get_time(&self, key: &str) -> Option<&chrono::NaiveDateTime> {
    self.times.get(key)
}
```

sseemayer: Oh sorry, I messed up the example code. Rust always requires explicit return type annotations for function definitions.

src/xml_parse.rs
jgrussell: As much as I wanted to include footgun-style design in those comments, I decided something a bit gentler might be better for publishing.

sseemayer (approved):

sseemayer:
Wonderful! 🎉  Thanks a lot for the nice contribution.

jgrussell:
Thank you again for the quick PR review and detailed feedback.

PR #43, opened 2021-08-31
Implement DB Metadata, Group UUID

### LINKED ISSUE #42: Implement metadata for KDBX4
(and possibly kdbx3/2/..)
Hi
I would like to be able to identify the trash bin (/recycle bin) group, I have a PR in progress that could use some.. heavy guidance on how to implement it best, but it is "working" for my project, on my usecase.

I'll submit it shortly

DavidVentura:
This fixes #42 
This commit implements a Metadata field for Database, whose only
member is 'recyclebin_uuid' (String).

It also implements a `uuid` field for Group, that can be matched with
the database's `meta.recyclebin_uuid`

I'll try to figure out how to get this tested as well. I don't know anything about the KDBX formats so I'm not sure if this also applies to v1 (which seems to not share the parser compared to v3/4?)


src/parse/kdbx3.rs
```diff
@@ -170,17 +170,24 @@ pub(crate) fn parse(data: &[u8], key_elements: &[Vec<u8>]) -> Result<Database> {
...
         times: Default::default(),
+        uuid: Default::default(),
     };
 
     // Parse XML data blocks
     for block_buffer in xml_blocks {
-        let block_group = xml_parse::parse_xml_block(&block_buffer, &mut *inner_decryptor)?;
+        let (block_group, _meta) =
+            xml_parse::parse_xml_block(&block_buffer, &mut *inner_decryptor)?;
+        // FIXME: This should only be done when _meta has a meaningful value
```
DavidVentura, line 189: not sure if this should check for `Default()` value or `parse_xml_block` should return an `Option`.
also not sure why here there are "blocks" when there's only 1 in v4

sseemayer:
Thanks a lot for your contribution, sorry for the wait in getting it merged!

PR #48, opened 2022-04-02
Fix #47: Concat binary attachment buffers before decompressing

### LINKED ISSUE #47: Reading KDBX files with (larger) attachments fail
**Description**
Files with larger attachments (2MB+) seem to be problematic when opened.

Here is a larger test file attached (needs unzip since GitHub doesn't allow attaching `kdbx` directly): [large-withfile.kdbx.zip](https://github.com/sseemayer/keepass-rs/files/7273982/large-withfile.kdbx.zip)
The password is `samplepassword`.

The reproducer file provided was provided by @Petemir, who created the [original report in keepass-diff](https://github.com/Narigo/keepass-diff/issues/35).

**Additional debug**
I have tried opening the attached file on the latest `master` with v0.4.9. Both debug commands fail with "corrupt deflate stream":

```
cargo run --features="utilities" --bin kp-dump-xml large-withfile.kdbx
Password (or blank for none): 
Error: IO { e: Custom { kind: InvalidInput, error: "corrupt deflate stream" } }

cargo run --features="utilities" --bin kp-show-db large-withfile.kdbx
Password (or blank for none): 
Error: IO { e: Custom { kind: InvalidInput, error: "corrupt deflate stream" } }
```

Adding debug `println!` statements, I could see that it [fails decompressing `block_buffer_compressed` in kdbx3 `parse`](https://github.com/sseemayer/keepass-rs/blob/96c805ab7b34e6ed9d72c6bf20bc8aa9b6d24998/src/parse/kdbx3.rs#L282).

DavidVentura:
This fixes #47.
I've left the signature of the function the same -- but it is not really useful as it always returns exactly 1 element

Seen here: https://github.com/Narigo/keepass-diff/issues/35


sseemayer:
Wonderful, thanks a lot for the fix! I am travelling at the moment and will try to make a crate release on Sunday.

sseemayer:
0.4.10 is released now.

PR #49, opened 2022-04-07
Updated dependecies and fixed issues related to them.

Hwatwasthat:
Could not update the AES dependency as it is affected by the deprecation of block-cipher, no easy replacement for ECB mode required for KDBX versions <4. Could have hacked it with a manual version but decided it was better to get the others up and running with the latest versions (which should bring some performance benefits on some platforms).

Results was changed due to unification of some results value upstream. Seems that there is less differentiation now on the key lengths etc. There is now a padding error that is returned, but should be careful about how that's exposed, as that way leads to oracle attacks if you're not careful. Not necessarily an issue for libraries but something to be born in mind for anything that uses this and returns data to users.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,23 +10,14 @@ checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
 
 [[package]]
 name = "aes"
-version = "0.7.4"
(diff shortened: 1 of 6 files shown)


sseemayer:
Thanks a lot for the contribution and sorry for the wait! I bumped all crate versions to the current ones.

PR #50, opened 2022-04-15
Make `Database::open` open database and key files

joezak11:


sseemayer:
Thanks for the PR! Can you comment on your rationale for making this change?

joezak11:
Since there is no need to open those files in the main code as I don't think there would any further use for it, it should be cleaner to let the `open` method handle it

sseemayer:
As the database and/or keyfile might come from another source than from disk, the currently-existing solution is more general, so I would like to keep it as-is.

PR #56, opened 2022-10-02
feat: add tags parsing

louib:
I added tags to the entry in the test database in order to test the changes.
The groups don't have tags, so no need to handle that case.


sseemayer:
wonderful, thank you! I will work on the PR backlog and also try to get back to your mail.

louib:
@sseemayer what happened to the git history on master?

PR #57, opened 2022-10-03
Bumps version of rust-argon2

sts10:
Github informed me of a security issue on [Medic](https://github.com/sts10/medic), namely "crossbeam-utils Race Condition vulnerability", linking to [this merge](https://github.com/crossbeam-rs/crossbeam/pull/781).
 
After some digging in `cargo tree`, I learned that keepass-xc depends on rust-argon2, which [depends on](https://github.com/sru-systems/rust-argon2/blob/master/Cargo.toml#L24) crossbeam-utils. 

Luckily, I think version 1.0 of rust-argon2 seems to use a safe version of crossbeam-utils ("0.8"). This PR simply bumps keepass-rs's rust-argon2 dependency to v 1.0. Tests pass.


sts10:
Ah, running `cargo tree` on my fork shows that rust-argon2 is _still_ using `crossbeam-utils v0.8.12`. My security warning on Medic says I need to use v0.8.7. 

I feel like I don't quite understand how Cargo handles dependency versions, since I thought `version = "0.8"` would automatically use v0.8.7.

Issue #14, opened 2019-10-07
Feature to encrypt a Keepass database


wusyong:
Really nice to have this useful lib in rust. However, it seems there's no way to save to a keepass file at the moment. Does it plan to add so?

sseemayer:
This would be a big amount of additional work. I am currently not planning to implement this myself but would be willing to accept PRs if anyone is willing to make progress on this.

louib:
Re-opening this issue until we have the feature available as a feature flag.

sseemayer:
#106 will make an experimental version of this available.

sseemayer:
Version 0.6.0-rc.1 is now out for testing on crates.io - experiences with the feature from a broader audience would be appreciated!

louib:
@sseemayer do you feel comfortable publishing version `0.6`? I don't think the `rc` versions have as much visibility, and the kdbx saving feature would still be behind a feature flag. I think it would generate more feedback if we publish a non-rc version.

sseemayer:
version `0.6.0` is now out on crates.io

sseemayer:
We have this available as a feature flag, so closing.

PR #58, opened 2022-10-05
Implement KDBX4 database saving

louib:
This is a first draft for KDBX4 database saving support. This is still a WIP, but I wanted to open the PR now in order to get early feedback. This PR also includes some refactorings that could be extracted to separate PRs in order to make the code review easier and safer.

Fixes https://github.com/sseemayer/keepass-rs/issues/14

## TODO
- [x] Implement AES `encrypt`
- [x] Implement TwoFish `encrypt`
- [x] Implement Salsa20 `encrypt`
- [x] Implement ChaCha20 `encrypt`
- [x] Implement GZip `compress`
- [x] Dump binary attachments
- [x] Implement protected attributes

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -1,13 +1,19 @@
 use flate2::read::GzDecoder;
-use std::io::Read;
+use flate2::write::GzEncoder;
+use flate2::Compression;
+use std::io::{Read, Write};
 
 pub trait Decompress {
+    fn compress(&self, in_buffer: &[u8]) -> Result<Vec<u8>, std::io::Error>;
     fn decompress(&self, in_buffer: &[u8]) -> Result<Vec<u8>, std::io::Error>;
 }
 
 pub struct NoCompression;
 
 impl Decompress for NoCompression {
+    fn compress(&self, in_buffer: &[u8]) -> Result<Vec<u8>, std::io::Error> {
+        Ok(in_buffer.to_vec())
+    }
     fn decompress(&self, in_buffer: &[u8]) -> Result<Vec<u8>, std::io::Error> {
         Ok(in_buffer.to_vec())
     }
@@ -16,6 +22,14 @@ impl Decompress for NoCompression {
 pub struct GZipCompression;
 
 impl Decompress for GZipCompression {
+    fn compress(&self, in_buffer: &[u8]) -> Result<Vec<u8>, std::io::Error> {
+        let mut res = Vec::new();
+        let mut encoder = GzEncoder::new(&mut res, Compression::default());
+        encoder.write_all(in_buffer)?;
+        encoder.flush()?;
+        encoder.finish()?;
+        Ok(res)
+    }
     fn decompress(&self, in_buffer: &[u8]) -> Result<Vec<u8>, std::io::Error> {
         let mut res = Vec::new();
         let mut decoder = GzDecoder::new(in_buffer);
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,11 +1,14 @@
 use hex_literal::hex;
+
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use thiserror::Error;
 
+use crate::crypt::ciphers::Cipher;
 use crate::{
+    compression,
     crypt::{ciphers, kdf, CryptographyError},
-    decompress,
-    variant_dictionary::VariantDictionary,
+    variant_dictionary::{VariantDictionary, VariantDictionaryValue},
 };
 
 const _CIPHERSUITE_AES128: [u8; 16] = hex!("61ab05a1946441c38d743a563df8dd35");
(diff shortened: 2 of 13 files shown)


varjolintu:
Nice work!

2franix:
Looks promising! I too would be interested to saving databases. Thanks for the addition!

sseemayer:
Since I made quite a mess of the git history in merging all the outstanding PRs, I have made a rebase of your changes onto the current master branch. You can find it at [kdbx4-saving](https://github.com/sseemayer/keepass-rs/tree/kdbx4-writing).

louib:
@sseemayer don't bother rebasing this PR. My plan is split this PR into smaller patches that are more focused and easier to review. For example, I'm going to create a PR that only extracts the constant values that are going to be used both when reading and writing. I also think that we're going to need to fence the database saving feature behind a feature flag. The reason is that writing a database with result in data loss until we support all the kdbx4 fields.

> Since I made quite a mess of the git history in merging all the outstanding PRs

I think we should enable branch protection on the master branch. We have many CI checks that should pass before any commit makes its way to the master branch, I don't see any reason not to leverage those CI jobs for every change to the codebase. What do you think?

sseemayer:
Sounds good! I have added something for the master branch.

louib:
This PR is ready for reviews. The `Database::save` function is only available from inside the library, using `pub(crate)`. This will give us more time to stabilize the new code and finish the XML parsing and dumping functions. As of right now, using the `Database::save` function **will** result in data loss because we do not parse and dump all the XML elements in a database.

I added a bunch of unit tests for many of the ciphers and kdfs combinations, but not all the combinations are covered. After this PR is merged, I will open a PR to switch to matrix testing so that we cover all the combinations. This will also reduce duplicated code in the tests.

Being able to parse and dump databases will be very helpful in the future to add more granular unit tests, because we won't have to rely on "magic" databases, committed to the repo as fixtures. It will still be useful to parse actual databases from other apps as integration tests, but at least that won't be our only option for testing.

I added a bunch of `TODO`s in the XML parsing functions for all the fields we are not parsing right now. Those fields were seen in a KeePassXC database. I think KeePassXC supports all the fields in the kdbx4 format, but I'm not 100% sure about that. 

@droidmonkey I'd like to have your opinion on [the default HMAC block size](https://github.com/sseemayer/keepass-rs/pull/58/files#diff-949b364ca9620e6a856467097dfd4aa149818bbc8acabf51a34dc57973b51a09R16). The block size (along with most of the hmac module) was inspired by the KeePassXC codebase. I'm honestly not sure if that value makes sense.

@DavidVentura I'd like to have your input on this PR, if you get a chance. The feature won't be available yet externally, so it won't be available to implement in [pykeepass-rs](https://github.com/DavidVentura/pykeepass-rs) right now, but I think we're pretty close to having the feature available at least behind a cargo feature flag.

src/compression.rs
```diff
@@ -1,13 +1,19 @@
 use flate2::read::GzDecoder;
-use std::io::Read;
+use flate2::write::GzEncoder;
+use flate2::Compression;
+use std::io::{Read, Write};
 
 pub trait Decompress {
```
sseemayer, line 6: We could rename the trait to `Compression` to make it clearer what it is for.

src/config.rs
```diff
@@ -211,7 +316,7 @@ impl TryFrom<VariantDictionary> for KdfSettings {
     }
 }
 
-#[derive(Debug)]
+#[derive(PartialEq, Debug)]
```
sseemayer, line 319: `PartialEq` might not be needed

sseemayer (approved):

sseemayer:
Thank you for the fantastic work, it looks great! 

I like the idea of getting things merged with a visibility of `pub(crate)` for now until we can do some more cleanups.

DavidVentura:
Hi!
I don't think I'll be able to review at least this week -- only think I'm able to contribute with is a similar effort I went through before:

[Add write support, #1](https://github.com/DavidVentura/libkeepass-rs/pull/1)

I'm not sure how helpful it'll be for you, I wasn't really sure what I was doing at that time, but it did seem to work!

src/compression.rs
louib: Maybe `CompressionAlgorithm`?

src/config.rs
louib: Addressed in https://github.com/sseemayer/keepass-rs/pull/85

louib:
@DavidVentura Thanks for linking the PR. I had a look at it, and there's a few things in there that I think we could port over here, notably all the `Group` and `Entry` fields that you implemented. We'll have to implement all of those and all the remaining `Database`/`Group`/`Entry` fields before making the new `Database::save` function.

droidmonkey:
> @droidmonkey I'd like to have your opinion on [the default HMAC block size](https://github.com/sseemayer/keepass-rs/pull/58/files#diff-949b364ca9620e6a856467097dfd4aa149818bbc8acabf51a34dc57973b51a09R16). The block size (along with most of the hmac module) was inspired by the KeePassXC codebase. I'm honestly not sure if that value makes sense.

I am personally not a fan of the data streams in the KeePassXC codebase, they are from KeePassX era. The block size is not relevant because it is embedded as part of the stream (HASH + BLOCKSIZE + DATA) so when you read the block it reads the size that was actually stored. So... the block size is really just for chunking the data out for efficient reads, 1 MiB seems reasonable to me 😄

louib:
@droidmonkey thanks for the review! Let's go with that default for now then.

PR #81, opened 2023-01-14
fix: pin the tarpaulin image and use stable rust

louib:
The current coverage workflow is failing with the latest tarpaulin image, as seen [here](https://github.com/sseemayer/keepass-rs/actions/runs/3912616535/jobs/6687480974).
This PR switches to the latest stable release instead. I tested locally, and I was able to generate the coverage report.

I think we'll also have to update the coverage badge in the README. The current codecov project page displays the following error message:
```
 This repo has been deactivated 
```

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -10,13 +10,13 @@ jobs:
     name: Code Coverage
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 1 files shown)


varjolintu (approved):

louib:
Looks like pushing a new code coverage report reactivated the project in codecov, so we don't have to update the badge in the README after all!

PR #84, opened 2023-01-15
Add basic otp support

DavidVentura:
Hi
I'll try to pull features off my fork into this one, not sure how applicable they are

* Add basic otp support
* Add bin for TOTP printing
* Return code with validity duration


src/lib.rs
```diff
@@ -10,6 +10,7 @@ mod keyfile;
 pub(crate) mod variant_dictionary;
 pub(crate) mod xml_parse;
 
+pub mod otp;
```
louib, line 13: @DavidVentura why `pub`? Couldn't it be `pub(crate)` or just `mod`?

src/otp.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::time::Duration;
+use std::time::{SystemTime, UNIX_EPOCH};
+use totp_lite::{totp_custom, Sha1, Sha256, Sha512};
+use url::Url;
+
+const DEFAULT_PERIOD: u64 = 30;
+const DEFAULT_DIGITS: u32 = 8;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum AlgoType {
```
louib, line 12: @DavidVentura how about `TOTPAlgorithm`? Or `TOTPCipher`.

Cargo.toml
```diff
@@ -43,6 +42,9 @@ chacha20 = "0.9"
 cipher = { version = "0.4", features = ["std"] }
 twofish = "0.7"
 cbc = "0.1"
+totp-lite = "2.0.0"
```
louib, line 45: @DavidVentura I feel like we should add a `totp` feature, so that those 3 dependencies are optional. That would allow for a more lightweight base library.

src/lib.rs
louib: If it's because of the tests, I would suggest moving the tests inside the `otp.rs` module, so that we don't have to expose the internal API only for testing.

louib (commented):
@DavidVentura thanks for the PR! There's a small conflict because of the kdbx4 PR, but shouldn't be too hard to rebase.

src/otp.rs
```diff
@@ -0,0 +1,116 @@
...
+            return None;
+        }
+        if issuer.is_none() {
+            return None;
+        }
+        Some(TOTP {
+            label,
+            secret: base32::decode(
+                base32::Alphabet::RFC4648 { padding: true },
+                &secret.unwrap(),
```
sseemayer, line 88: Instead of `unwrap`ping here, you could use `secret?` to either use the value inside the `Option`, or do an early `return None`. This also gets rid of lines 78-83.

src/lib.rs
DavidVentura: I think it was because I needed access from the Python bindings, but I'm not sure -- regardless, it's not relevant for this change, I'll remove it

sseemayer:
Thank you very much for the contribution!

PR #88, opened 2023-01-17
feat: use argon2 kdf settings

louib:
When I modified the database creation function to actually use the argon2 settings, I started seeing the following test errors:
```
thread 'xml_parse::xml_tests::test_entry' has overflowed its stack
```

Looks like there's a minimum memory required to derive the hashed password with argon2. I'm wondering if we should provide a default value, and a minimum value, to make sure that the value is large enough. I didn't see any recommendations on the crate's documentation.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -415,22 +415,28 @@ impl Database {
         master_seed.resize(crate::parse::kdbx4::HEADER_MASTER_SEED_SIZE.into(), 0);
         getrandom::getrandom(&mut master_seed)?;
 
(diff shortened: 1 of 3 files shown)


sseemayer:
[According to the argon2 crate docs](https://docs.rs/argon2/latest/argon2/struct.Params.html#associatedconstant.DEFAULT_M_COST), memory cost is specified in kilobytes (which might explain the errors) and defaults to 4096 kilobytes.

sseemayer (approved):

PR #89, opened 2023-01-17
fix: Prevent stack overflow due to circular type conversions of nested error types

DavidVentura:
Hi
Only adding a test because I don't understand what the problem is
When working on the TOTP PR, I noticed that providing a bad password gave:
```
fatal runtime error: stack overflow
```

I've isolated it to this `Error` type conversion, but the `#Error` macro is magic

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -63,6 +63,9 @@ pub enum DatabaseKeyError {
     IncorrectKey,
 
     #[error(transparent)]
+    Cryptography(#[from] CryptographyError),
+
+    #[error(transparent)]
     Keyfile(#[from] KeyfileError),
 }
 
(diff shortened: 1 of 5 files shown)


sseemayer:
The issue was due to a circular type conversion. `From<DatabaseKeyError> for DatabaseOpenError` was called, which was implemented as:

```rust
impl From<DatabaseKeyError> for DatabaseOpenError {
    fn from(value: DatabaseKeyError) -> Self {
        value.into()
    }
}
```
the implicit `.into()` was resolved to calling `DatabaseOpenError::from::<DatabaseKeyError>(value)`, leading to an infinite recursion and, eventually, the stack overflow.

I've added another commit that makes the type conversions between error types more explicit and also narrows the scope of some function's error return type.

louib (approved):
Awesome team effort!

PR #93, opened 2023-01-18
ci: add a workflow to lint commits

louib:
The default help URL will be [this one](https://github.com/conventional-changelog/commitlint/#what-is-commitlint), but we can change it if we find it not helpful enough. For example, we could redirect to our own `CONTRIBUTING.md` once we have one.
This is required before we can add a workflow that automates the releases after a merge on master.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,11 @@
+name: Lint Commit Messages
+on: [pull_request]
+
+jobs:
+  commitlint:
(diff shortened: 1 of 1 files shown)


sseemayer (approved):

louib:
@sseemayer how do you feel about making the new workflow required as well? The other workflows are also all required.

sseemayer:
Good idea, especially in preparation of automated releases - it's now also required.

PR #97, opened 2023-01-21
refactor(xml)!: split into modules

sseemayer:
Work on the `xml_parse` module to split parsing and dumping code into separate submodules and reduce repetitive code, in preparation for adding additional fields from the XML format.

- [x] split out dumping
- [x] refactor dumping to use a shared trait
- [x] split out parsing
- [x] refactor parsing to use a shared trait


sseemayer:
@louib @varjolintu Sorry for the large changeset. I hope that this architecture for the XML parsing/dumping code will make it much easier to implement the missing fields to the XML document.

src/parse/kdbx3.rs
```diff
@@ -190,16 +190,15 @@ pub(crate) fn parse(data: &[u8], key_elements: &[Vec<u8>]) -> Result<Database, D
...
         name: "Root".to_owned(),
         children: Default::default(),
-        expires: Default::default(),
         times: Default::default(),
     };
 
     // Parse XML data blocks
     for block_buffer in xml_blocks {
-        let (block_group, _meta) = xml_parse::parse_xml_block(&block_buffer, &mut *inner_decryptor)
+        let kxml = crate::xml_db::parse::parse(&block_buffer, &mut *inner_decryptor)
```
louib, line 198: @sseemayer how about `raw_database`, or `database_content`?

src/parse/kdbx4.rs
```diff
@@ -387,13 +387,13 @@ pub(crate) fn parse(data: &[u8], key_elements: &[Vec<u8>]) -> Result<Database, D
         .inner_random_stream
         .get_cipher(&inner_header.inner_random_stream_key)?;
 
-    let (root, meta) = xml_parse::parse_xml_block(&xml, &mut *inner_decryptor)?;
+    let kpxml = crate::xml_db::parse::parse(&xml, &mut *inner_decryptor)?;
```
louib, line 390: @sseemayer same comment as for `kxml`: I think we could use a more explicit name here.

src/xml_db/parse/entry.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        Ok(out)
+    }
+}
+
+struct History {
+    entries: Vec<Entry>,
+}
+
+impl FromXml for History {
```
louib, line 430: @sseemayer is this implemented? I don't see it added to the `Entry`.
If we don't implement the entry history in this PR, I would suggest adding this to another PR, to make this one easier to review.

tests/entry_tests.rs
```diff
@@ -16,7 +16,7 @@ mod tests {
             assert_eq!(e.get_url(), Some("http://keepass.info/"));
             assert_eq!(e.get("custom attribute"), Some("data for custom attribute"));
             assert_eq!(e.get("URL"), Some("http://keepass.info/"));
-            assert_eq!(e.expires, false);
+            assert_eq!(e.times.expires, false);
```
louib, line 19: @sseemayer this is a breaking change. Ideally we would mark the commit message with `BREAKING CHANGE`, but now that I read the [semantic-release docs](https://github.com/semantic-release/semantic-release#commit-message-format),it looks like this would bump the major version. This is not what we want, at least not until we reach v1.
I think we can skip the `BREAKING CHANGE` for now, but we should make it clear in the commit message that
the field was moved.

louib (commented):
Nice cleanup! Added a few comments.

src/parse/kdbx3.rs
sseemayer: agreed and changed to `database_content`.

src/parse/kdbx4.rs
sseemayer: agreed and changed to `database_content`.

src/xml_db/parse/entry.rs
sseemayer: With the current architecture, I essentially had to implement all the XML parsers or run into errors when parsing the XML document. The only part missing is adding fields to the appropriate `Entry`, `Group`, `Meta`, etc. fields (another breaking change), and then assigning the results of the XML parsing to those fields.

tests/entry_tests.rs
louib: If you manually release `0.6.0`, then I think we are good. Otherwise semantic-release would pick up all the commit since the last release and create a release with that, and we would get a `1.0.0` because of the breaking change.

src/xml_db/parse/entry.rs
```diff
@@ -0,0 +1,481 @@
...
+                    }
+                    _ => {
+                        return Err(XmlParseError::BadEvent {
+                            expected: "valid Entry child",
+                            event: event.clone(),
+                        })
+                    }
+                },
+                SimpleXmlEvent::End(name) if name == "Entry" => break,
+                _ => {
```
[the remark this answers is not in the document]
sseemayer: Added the `IgnoreSubfield` parser that can be used to ignore unknown fields.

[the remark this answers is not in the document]
louib: @sseemayer is it already used? Would be nice to add a unit test to make sure we don't throw an error in case of unknown field. We could also add an option in the future to enable `strict` parsing, otherwise we always risk losing information when saving if the field is not supported.

PR #99, opened 2023-01-28
feat!: parse and dump all Meta fields

sseemayer:
Implement reading and writing of all database metadata.

BREAKING CHANGE: `recyclebin_uuid` is now an `Option<String>`

### DIFF
--- a/src/parse/kdbx4.rs
+++ b/src/parse/kdbx4.rs
@@ -1,6 +1,7 @@
 use std::convert::TryFrom;
 use std::convert::TryInto;
 
+use crate::meta::BinaryAttachment;
 use crate::{
     config::{Compression, InnerCipherSuite, KdfSettings, OuterCipherSuite},
     crypt,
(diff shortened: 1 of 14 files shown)


sseemayer:
I am aware that the test coverage is decreasing, but haven't figured out a good way to test the XML parser errors (the primary parts that are uncovered). Any suggestions for how to do so somewhat elegantly?

src/parse/kdbx4.rs
```diff
@@ -1,6 +1,7 @@
 use std::convert::TryFrom;
 use std::convert::TryInto;
 
+use crate::meta::BinaryAttachment;
```
louib, line 4: @sseemayer nitpicking here, but why not merge that import with the import block below?

src/xml_db/dump/mod.rs
```diff
@@ -163,137 +150,63 @@ impl DumpXml for Database {
...
-        // TODO RecycleBinChanged
-        // TODO EntryTemplatesGroup
-        // TODO EntryTemplatesGroupChanged
-        // TODO LastSelectedGroup
-        // TODO LastTopVisibleGroup
-        // TODO HistoryMaxItems
-        // TODO HistoryMaxSize
-        // TODO SettingsChanged
-        // TODO CustomData
+        writer.write(WriterEvent::start_element("Times"))?;
```
louib, line 159: @sseemayer at some point I think we should extract the xml tag names as constants, since they are repeated in multiple places in the codebase.

louib (approved):
Great work! :rocket:

louib:
> Any suggestions for how to do so somewhat elegantly?

We could simply feed crafted XML dumps to `parse` with an `inner_cipher` that is a `PlainCipher`. That would make the test cases very explicit and allow us to test pretty much everything.

PR #102, opened 2023-01-29
feat!: implement remaining XML fields

sseemayer:
Now that we have a solid testing framework and the parsing and dumping infrastructure has matured, hammer out the final missing fields from the XML database

- [x] Parse and dump `Group` fields
- [x] Parse and dump `Entry` fields

BREAKING CHANGE: Adds additional fields to Group and Entry

### DIFF
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -9,7 +9,7 @@ use crate::db::Times;
 use crate::otp::{TOTPError, TOTP};
 
 /// A database entry containing several key-value fields.
-#[derive(Debug, Default, Eq, PartialEq)]
+#[derive(Debug, Default, Eq, PartialEq, Clone)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct Entry {
     pub uuid: String,
@@ -136,7 +136,7 @@ impl serde::Serialize for Value {
 }
 
 /// An AutoType setting associated with an Entry
-#[derive(Debug, Default, Eq, PartialEq)]
+#[derive(Debug, Default, Eq, PartialEq, Clone)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct AutoType {
     pub enabled: bool,
@@ -145,7 +145,7 @@ pub struct AutoType {
 }
 
 /// A window association associated with an AutoType setting
-#[derive(Debug, Default, Eq, PartialEq)]
+#[derive(Debug, Default, Eq, PartialEq, Clone)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct AutoTypeAssociation {
     pub window: Option<String>,
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use secstr::SecStr;
 use uuid::Uuid;
 
-use crate::db::Times;
+use crate::{db::Times, CustomData};
 
 #[cfg(feature = "totp")]
 use crate::otp::{TOTPError, TOTP};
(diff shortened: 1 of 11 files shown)


sseemayer:
While this should make sure that almost no data from a loaded kdbx4 file is not lost when writing, a lot of the APIs are still pretty ugly and there is some duplication (e.g. `BinaryAttachment`) because of different ways for handling fields between the KDBX versions.

louib:
> While this should make sure that almost no data from a loaded kdbx4 file is lost when writing, a lot of the APIs are still pretty ugly and there is some duplication (e.g. BinaryAttachment) because of different ways for handling fields between the KDBX versions.

Do you want to wait before rolling out the `save_kp4` feature? I feel comfortable shipping the feature if we disable it by default and add a disclaimer to the README.

src/db/entry.rs
```diff
@@ -18,15 +18,25 @@ pub struct Entry {
     pub tags: Vec<String>,
 
     pub times: Times,
+
+    pub custom_data: CustomData,
+
+    pub icon_id: Option<usize>,
+    pub custom_icon_uuid: Option<String>,
```
louib, line 25: @sseemayer this could be refactored into a single enum, since those should never be used at the same time I believe.

src/db/entry.rs
```diff
@@ -18,15 +18,25 @@ pub struct Entry {
...
 }
 impl Entry {
     pub fn new() -> Entry {
         Entry {
             uuid: Uuid::new_v4().to_string(),
-            fields: HashMap::default(),
-            times: Times::default(),
-            autotype: None,
-            tags: vec![],
+            ..Default::default()
```
louib, line 39: @sseemayer TIL!

src/db/group.rs
```diff
@@ -17,20 +17,44 @@ pub struct Group {
...
+
+    /// Whether autotype is enabled
+    // TODO: in example XML files, this is "null" - what should the type be?
+    pub enable_autotype: Option<String>,
+
+    /// Whether searching is enabled
+    // TODO: in example XML files, this is "null" - what should the type be?
+    pub enable_searching: Option<String>,
+
+    pub last_top_visible_entry: Option<String>,
```
louib, line 49: @sseemayer I'm guessing this is an entry's UUID? We could probably document that.

src/xml_db/mod.rs
```diff
@@ -105,6 +135,26 @@ mod tests {
...
+        subgroup.custom_icon_uuid = Some("CUSTOM-ICON".to_string());
+        subgroup.times.expires = true;
+        subgroup.times.usage_count = 100;
+        subgroup
+            .times
+            .times
+            .insert("Created".to_string(), NaiveDateTime::default());
+        subgroup.is_expanded = true;
+        subgroup.default_autotype_sequence =
+            Some("{UP}{UP}{DOWN}{DOWN}{LEFT}{RIGHT}{LEFT}{RIGHT}BA".to_string());
```
louib, line 150: @sseemayer :laughing:

src/xml_db/parse/entry.rs
```diff
@@ -42,7 +43,6 @@ impl FromXml for Entry {
                                 .split(|c| c == ';' || c == ',')
                                 .map(|x| x.to_owned())
                                 .collect();
-                            out.tags.sort();
```
louib, line 45: @sseemayer I ran into problems because of that line in the past as well. I agree that we should remove it.

louib (approved):
Great job! The cleanups are paying off, this PR was super easy to review

src/db/group.rs
sseemayer: That item was previously intentionally left undocumented because I had no clue what "last top visible" is supposed to mean. It is the UUID of some entry, but even going back into the original KeePass sourcecode, it is not 100% clear to me what it is about - something about restoring selected item states when re-opening a database. 

I will add some documentation text that clarifies that it is supposed to be an item UUID, but won't bother to reverse engineer this field further for the time being.

src/db/entry.rs
sseemayer: I thought so too, but according to the KeePass sourcecode (which I cannot link into because it is served in a zip file from SourceForge, not even from subversion 😱 , see e.g. `KeePassLib/Serialization/KdbxFile.Write.cs:511`), `<IconId>` will always be present, and `<CustomIconUuid>` will be optionally present.

sseemayer:
> > While this should make sure that almost no data from a loaded kdbx4 file is lost when writing, a lot of the APIs are still pretty ugly and there is some duplication (e.g. BinaryAttachment) because of different ways for handling fields between the KDBX versions.
> 
> Do you want to wait before rolling out the `save_kp4` feature? I feel comfortable shipping the feature if we disable it by default and add a disclaimer to the README.

I agree but would put this into a separate PR.

src/db/entry.rs
louib: > which I cannot link into because it is served in a zip file from SourceForge, not even from subversion

:scream: indeed

PR #106, opened 2023-01-30
feat!: expose KDBX4 saving behind feature flag

sseemayer:
- [x] Create feature flag and change visibility
- [x] Update docs with disclaimers
- [x] Fix issues with opening created files
- [x] Test drive the feature somewhat to prevent catastrophic data losses

### DIFF
--- a/src/variant_dictionary.rs
+++ b/src/variant_dictionary.rs
@@ -28,6 +28,9 @@ pub enum VariantDictionaryError {
 
     #[error("Mistyped value: {}", key)]
     Mistyped { key: String },
+
+    #[error("VariantDictionary did not end with null byte, when it should")]
+    NotTerminated,
 }
 
 impl VariantDictionary {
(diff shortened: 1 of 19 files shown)


sseemayer:
Right now, all tests pass but the generated database file from the readme example cannot be opened by KeePassXC, so more testing is definitely needed.

sseemayer:
Made some progress by stepping through the KeePassXC sourcecode - at the end of `VariantDictionary` dumping, there needs to be a null byte. I am now running into errors parsing the inner header.

src/variant_dictionary.rs
```diff
@@ -141,6 +141,9 @@ impl VariantDictionary {
             data.extend_from_slice(&field_buffer);
         }
 
+        // signify end of variant dictionary
+        data.push(0);
```
louib, line 145: @sseemayer does that mean that we were also not verifying that when parsing? Should we add it?

sseemayer: Yes, we should. There will be probably more details like this that will be uncovered before files generated by us can be opened by other KeePass implementations.

sseemayer: resolved with da4c9d5 .

sseemayer:
I can now use the project to parse my personal KeePass database, dump it again and re-open it in KeePassXC. From diff-ing the inner XML file, I know that some parts are still missing such as DeletedObjects, but putting it behind an experimental feature flag makes sense so we can gather some additional user feedback.

PR #115, opened 2023-02-05
refactor: reduce allocations

sseemayer:
Make the dumping and variant dictionary code more idiomatic by reducing the number of clones and temporary allocations.

Changes the code to use `std::io::Write` instead of returning `Vec<u8>` and to use more references instead of cloning owned values.


src/db/mod.rs
```diff
@@ -360,22 +367,9 @@ impl Database {
...
-                return Err(DatabaseSaveError::UnsupportedVersion.into());
-            }
-            Header::KDBX4(_) => {
-                self.generate_ivs()?;
-                crate::format::kdbx4::dump(self, &key_elements)
-            }
-        };
-
-        encrypted_db
+        let mut data = Vec::new();
```
louib, line 370: @sseemayer ah! That's what I was looking for yesterday, I didn't realize that you could use a simple `Vec` as an `std::io::Write`! I don't think the `dump` function is relevant anymore in that case.

louib:
@sseemayer very nice improvement!

PR #116, opened 2023-02-05
refactor!: never store keys or seeds

sseemayer:
Instead of overwriting the various key seeds and IVs and testing that they were overwritten, only keep the buffers around during parsing and dynamically regenerate them during dumping.

BREAKING CHANGE: Database interface changed so that headers are not exposed directly anymore, but instead only DatabaseSettings are.

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -27,10 +27,11 @@ use crate::{
         KdfSettingsError, OuterCipherSuite, OuterCipherSuiteError,
     },
     crypt::{calculate_sha256, CryptographyError},
+    db::meta::BinaryAttachments,
     format::{
-        kdb::{parse_kdb, KDBHeader},
-        kdbx3::{parse_kdbx3, KDBX3Header},
-        kdbx4::{parse_kdbx4, KDBX4Header, KDBX4InnerHeader},
+        kdb::parse_kdb,
+        kdbx3::{decrypt_kdbx3, parse_kdbx3},
+        kdbx4::{decrypt_kdbx4, parse_kdbx4},
         DatabaseVersion, KDBX4_CURRENT_MINOR_VERSION,
     },
     hmac_block_stream::BlockStreamError,
@@ -39,30 +40,15 @@ use crate::{
     xml_db::parse::XmlParseError,
 };
 
-#[derive(Debug)]
-pub enum Header {
-    KDB(KDBHeader),
-    KDBX3(KDBX3Header),
-    KDBX4(KDBX4Header),
-}
-
-#[derive(Debug)]
-pub enum InnerHeader {
-    None,
-    KDBX4(KDBX4InnerHeader),
-}
-
 /// A decrypted KeePass database
 #[derive(Debug)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct Database {
-    /// Header information of the KeePass database
-    #[cfg_attr(feature = "serialization", serde(skip))]
-    pub header: Header,
+    /// Settings of the database such as encryption and compression algorithms
+    pub settings: DatabaseSettings,
 
-    /// Optional inner header information
-    #[cfg_attr(feature = "serialization", serde(skip))]
-    pub inner_header: InnerHeader,
+    /// Binary attachments in the inner header
+    pub header_attachments: BinaryAttachments,
 
     /// Root node of the KeePass database
     pub root: Group,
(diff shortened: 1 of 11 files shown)


src/db/mod.rs
```diff
@@ -344,26 +333,26 @@ impl Database {
     /// Save a database to a std::io::Write
     #[cfg(feature = "save_kdbx4")]
     pub fn save(
-        &mut self,
```
louib, line 347: @sseemayer good stuff!

src/db/mod.rs
```diff
@@ -394,11 +383,11 @@ impl Database {
     }
 
     /// Helper function to load a database into its internal XML chunks
-    pub fn get_xml_chunks(
+    pub fn get_xml(
```
louib, line 386: @sseemayer I like this renaming. Good idea to do those before the API becomes stable with `1.X`

src/format/kdbx4/dump.rs
```diff
@@ -42,38 +66,47 @@ pub fn dump_kdbx4(
...
+        .get_compression()
+        .compress(&payload)?;
 
-    let payload_encrypted = header
-        .outer_cipher
-        .get_cipher(&master_key, header.outer_iv.as_ref())?
+    let payload_encrypted = db
+        .settings
+        .outer_cipher_suite
+        .get_cipher(&master_key, &outer_iv)?
```
louib, line 109: @sseemayer could we move the IV generation into the `get_cipher` directly? I think this would make it even less error-prone. We could have a `get_cipher` that takes an existing IV (used for parsing), and a `get_fresh_cipher` that only takes a master key and that returns the cipher and a new IV (used for saving).

src/format/kdbx4/dump.rs
```diff
@@ -121,15 +154,19 @@ impl KDBX4Header {
...
+    ) -> Result<(), DatabaseSaveError> {
         writer.write(&[INNER_HEADER_RANDOM_STREAM_ID])?;
         writer.write_u32::<LittleEndian>(4)?;
         writer.write_u32::<LittleEndian>(self.inner_random_stream.dump())?;
 
         writer.write_u8(INNER_HEADER_RANDOM_STREAM_KEY)?;
         writer.write_with_len(&self.inner_random_stream_key)?;
 
-        for binary in &self.binaries {
+        for binary in &binaries.binaries {
```
louib, line 169: @sseemayer maybe the parameter could be called `attachments` instead? This would make the current referencing less awkward.

src/format/kdbx4/mod.rs
```diff
@@ -66,11 +61,12 @@ mod kdbx4_tests {
         inner_cipher_suite: InnerCipherSuite,
         kdf_setting: KdfSettings,
     ) {
-        let mut db = Database::new(NewDatabaseSettings {
+        let mut db = Database::new(DatabaseSettings {
+            version: DatabaseVersion::KDB4(0),
```
louib, line 65: @sseemayer we have a constant value for the inner version number, called `KDBX4_CURRENT_MINOR_VERSION`

src/format/kdbx4/parse.rs
```diff
@@ -79,16 +85,21 @@ pub(crate) fn decrypt_kdbx4(
     // derive master key from composite key, transform_seed, transform_rounds and master_seed
     let key_elements: Vec<&[u8]> = key_elements.iter().map(|v| &v[..]).collect();
     let composite_key = crypt::calculate_sha256(&key_elements)?;
-    let transformed_key = header.kdf.get_kdf().transform_key(&composite_key)?;
-    let master_key = crypt::calculate_sha256(&[header.master_seed.as_ref(), &transformed_key])?;
+    let transformed_key = outer_header
+        .kdf_settings
+        .get_kdf(&outer_header.kdf_seed)
```
louib, line 90: @sseemayer same comment as with `get_cipher`: We could have a `get_kdf` for parse, and a `get_fresh_kdf` that would encapsulate the IV generation.

louib (approved):

src/format/kdbx4/dump.rs
sseemayer: This should be addressed with #120

src/format/kdbx4/mod.rs
sseemayer: Right! I did use this in another location but must have missed this one. #121 should fix it.

src/format/kdbx4/parse.rs
sseemayer: Good idea, added in #123

src/format/kdbx4/dump.rs
sseemayer: The issue with this is that the `master_key` takes additional steps and inputs to derive, which would make the moved IV generation more complicated.

PR #120, opened 2023-02-05
refactor!: split BinaryAttachment and HeaderAttachment

### LINKED ISSUE #117: Split BinaryAttachments
Binary attachments can be provided both in the inner header of a KDBX4 file, and also as a `<Binaries>` tag inside of the `<Meta>` tag (this probably is only used in KDBX version 3). Both are currently handled with the `BinaryAttachments` and `BinaryAttachment` types, which is not a clean way to handle the data.

Split into separate types for attachments that are provided in the header and in the XML part of the file.

sseemayer:
Closes #117

Since the fields for an attachment inside of a <Meta> tag and inside of a KDBX4 binary header are different, make the distinction explicit by splitting up the types

BREAKING CHANGE: New type for header_attachments field in Database
BREAKING CHANGE: Fewer members for BinaryAttachment parsed from XML

### DIFF
--- a/src/db/meta.rs
+++ b/src/db/meta.rs
@@ -101,7 +101,6 @@ pub struct BinaryAttachments {
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct BinaryAttachment {
     pub identifier: Option<String>,
-    pub flags: u8,
     pub compressed: bool,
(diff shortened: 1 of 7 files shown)


louib (commented):
:heavy_check_mark:

PR #123, opened 2023-02-05
refactor: add KDF seed generation to KdfSettings

sseemayer:
See https://github.com/sseemayer/keepass-rs/pull/116#discussion_r1096788746

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -207,7 +207,21 @@ impl KdfSettings {
         }
     }
 
-    pub(crate) fn get_kdf(&self, seed: &[u8]) -> Box<dyn kdf::Kdf> {
+    /// For writing out a database, generate a new KDF seed from the settings and return the KDF
+    /// and the generated seed
+    pub(crate) fn get_kdf_and_seed(
+        &self,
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -207,7 +207,21 @@ impl KdfSettings {
         }
     }
 
-    pub(crate) fn get_kdf(&self, seed: &[u8]) -> Box<dyn kdf::Kdf> {
+    /// For writing out a database, generate a new KDF seed from the settings and return the KDF
+    /// and the generated seed
+    pub(crate) fn get_kdf_and_seed(
+        &self,
+    ) -> Result<(Box<dyn kdf::Kdf>, Vec<u8>), getrandom::Error> {
+        let mut kdf_seed = vec![0; self.seed_size()];
```
louib, line 215: @sseemayer I think we can make the `seed_size` function private now.

louib (approved):

src/config.rs
sseemayer: We could. Before going 1.0, I would want to look at the complete public interface and check what can be hidden.

PR #127, opened 2023-02-06
refactor!: clean up public interface

sseemayer:
This commit minimizes the public interface of the crate, and restructures it to make the docs more legible.

BREAKING CHANGE: Many visibility changes.

BREAKING CHANGE: Renamed and moved several types around.

BREAKING CHANGE: New Key type used to specify password and keyfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,22 +11,22 @@ Rust KeePass database file parser for KDB, KDBX3 and KDBX4, with experimental su
 
 ## Example
 ```rust
+use anyhow::Result;
 use keepass::{
-    db::{Database, NodeRef},
-    error::DatabaseOpenError,
-    Key
+    db::NodeRef,
+    Database, DatabaseKey
 };
 use std::fs::File;
 
-fn main() -> Result<(), DatabaseOpenError> {
+fn main() -> Result<()> {
     // Open KeePass database
     let path = std::path::Path::new("tests/resources/test_db_with_password.kdbx");
     let db = Database::open(
-        &mut File::open(path)?,         // the database
-        Key::with_password("demopass"), // password (keyfile is also supported)
+        &mut File::open(path)?,         		// the database
+        DatabaseKey::with_password("demopass"), // password (keyfile is also supported)
     )?;
 
-    // Iterate over all Groups and Nodes
+    // Iterate over all `Group`s and `Entry`s
     for node in &db.root {
         match node {
             NodeRef::Group(g) => {
@@ -85,14 +85,13 @@ You can enable the experimental support for saving KDBX4 databases using the `sa
 ```rust ignore
 use anyhow::Result;
 use keepass::{
-    db::{Database, DatabaseSettings, Entry, Group, Node, NodeRef, Value},
-    error::DatabaseOpenError,
-    Key
+    db::{Database, Entry, Group, Node, NodeRef, Value},
+    DatabaseKey
 };
 use std::fs::File;
 
 fn main() -> Result<()> {
-    let mut db = Database::new(DatabaseSettings::default())?;
+    let mut db = Database::new(Default::default())?;
 
     db.meta.database_name = Some("Demo database".to_string());
 
@@ -109,7 +108,7 @@ fn main() -> Result<()> {
 
     db.save(
         &mut File::create("demo.kdbx")?,
-        Key::with_password("demopass"),
+        DatabaseKey::with_password("demopass"),
     )?;
 
     Ok(())
--- a/README.md
+++ b/README.md
@@ -11,7 +11,11 @@ Rust KeePass database file parser for KDB, KDBX3 and KDBX4, with experimental su
 
 ## Example
 ```rust
-use keepass::{Database, DatabaseOpenError, NodeRef};
+use keepass::{
+    db::{Database, NodeRef},
+    error::DatabaseOpenError,
+    Key
+};
 use std::fs::File;
 
 fn main() -> Result<(), DatabaseOpenError> {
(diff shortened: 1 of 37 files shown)


README.md
```diff
@@ -11,16 +11,19 @@ Rust KeePass database file parser for KDB, KDBX3 and KDBX4, with experimental su
 
 ## Example
 ```rust
-use keepass::{Database, DatabaseOpenError, NodeRef};
+use keepass::{
+    db::{Database, NodeRef},
+    error::DatabaseOpenError,
+    Key
```
louib, line 17: @sseemayer how about `DatabaseKey`? That would be consistent with the other public struct we expose.

src/config.rs
```diff
@@ -20,141 +20,129 @@ const PLAIN: u32 = 0;
...
+pub enum OuterCipherConfig {
     AES256,
     Twofish,
     ChaCha20,
 }
 
+/// Errors for outer encryption configuration
 #[derive(Debug, Error)]
-pub enum OuterCipherSuiteError {
+pub enum OuterCipherConfigError {
```
louib, line 34: @sseemayer could be only `OuterCipherError`?

src/config.rs
```diff
@@ -20,141 +20,129 @@ const PLAIN: u32 = 0;
...
+pub enum InnerCipherConfig {
     Plain,
     Salsa20,
     ChaCha20,
 }
 
+/// Errors with inner protected value encryption
 #[derive(Debug, Error)]
-pub enum InnerCipherSuiteError {
+pub enum InnerCipherConfigError {
```
louib, line 100: @sseemayer same comment as above, could be `InnerCipherError`

src/db/meta.rs
```diff
@@ -38,65 +40,106 @@ pub struct Meta {
...
+    /// last time the recycle bin was changed
     pub recyclebin_changed: Option<NaiveDateTime>,
 
+    /// UUID of the group containing entry templates
     pub entry_templates_group: Option<String>,
 
+    /// last time the group containing entry templates was changed
     pub entry_templates_group_changed: Option<NaiveDateTime>,
 
+    /// last selected group UUID
```
louib, line 64: @sseemayer to keep the same format as `entry_templates_group`:
```
UUID of the last selected group
```

src/db/meta.rs
```diff
@@ -38,65 +40,106 @@ pub struct Meta {
...
+    /// Maximum number of items of history to keep
     pub history_max_items: Option<usize>,
 
+    /// Maximum size of the history to keep
     pub history_max_size: Option<usize>,
 
+    /// Last time the settings were changed
     pub settings_changed: Option<NaiveDateTime>,
 
+    /// Binary attachments in the Metadata header
```
louib, line 78: @sseemayer do we have to mention that they are store in the Metadata header? That feels like an implementation detail. We could simply say `Binary attachments stored in the database`.

src/db/mod.rs
```diff
@@ -56,230 +51,35 @@ pub struct Database {
...
-impl From<CompressionError> for DatabaseOpenError {
-    fn from(e: CompressionError) -> Self {
-        DatabaseIntegrityError::from(e).into()
-    }
-}
-
+/// Settings for how a database is stored on disk
 #[derive(Debug)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct DatabaseSettings {
```
louib, line 57: @sseemayer should that not be `DatabaseConfig`? To be consistent with the other configuration structs.

src/db/mod.rs
```diff
@@ -386,7 +146,7 @@ impl Database {
         Ok(data)
     }
 
-    /// Get the version of a database.
+    /// Get the version of a database without loading it
```
louib, line 149: @sseemayer `without decrypted it`? We load it but just a bit :laughing:

src/error.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// A cryptography error occurred while writing the database
+    #[error(transparent)]
+    Cryptography(#[from] CryptographyError),
+
+    /// An error getting randomness for keys occurred
+    #[error(transparent)]
+    Random(#[from] getrandom::Error),
+}
+
+// move error type conversions to a module and exclude them from coverage counting.
```
louib, line 180: @sseemayer good call!

louib (approved):
Great job! I added a few comments, but no blockers.

src/db/mod.rs
sseemayer: I've cleaned it up in a follow-up commit.

src/db/mod.rs
sseemayer: Good point, I've renamed this and a bunch of other types to have a consistent naming.

src/db/meta.rs
sseemayer: I would like to keep it for now, since there are two different ways of specifying binary attachments in the file format, and users might be confused about that. The longer-term goal would be to consolidate into one attachment field.

src/db/meta.rs
sseemayer: Agreed and changed in a follow-up commit.

src/config.rs
sseemayer: Since the error is not with the cipher itself, but with its configuration, I would prefer to keep the name as-is.

src/config.rs
sseemayer: Since the error is not with the cipher itself, but with its configuration, I would prefer to keep the name as-is.

README.md
sseemayer: Agreed and adjusted accordingly.

PR #133, opened 2023-02-11
feat!: parse base64-encoded UUID values

### LINKED ISSUE #131: Better UUID handling
We are currently keeping UUIDs from the XML document as Base64-encoded strings, but could parse them into proper `Uuid` objects. Also, [UUIDs generated by us might not be base64 encoded](https://github.com/sseemayer/keepass-rs/blob/e73c55a47d243f52be028cf4a64d5a6d62e27d05/src/db/entry.rs#L38), leading to incorrect XML output.

sseemayer:
Closes #131

Instead of keeping the UUIDs of objects around as base64-encoded strings, actually parse them into Uuid objects. This also fixes some issues where UUID strings instead of base64-encoded UUIDs were serialized into the XML files.

BREAKING CHANGE: Changed type of UUID values from String to `uuid::Uuid`

### DIFF
--- a/src/db/group.rs
+++ b/src/db/group.rs
@@ -12,7 +12,7 @@ use crate::db::{
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct Group {
     /// The unique identifier of the group
-    pub uuid: String,
+    pub uuid: Uuid,
 
     /// The name of the group
     pub name: String,
@@ -24,7 +24,7 @@ pub struct Group {
     pub icon_id: Option<usize>,
 
     /// UUID for a custom group icon
-    pub custom_icon_uuid: Option<String>,
+    pub custom_icon_uuid: Option<Uuid>,
 
     /// The list of child nodes (Groups or Entries)
     pub children: Vec<Node>,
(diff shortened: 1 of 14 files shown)


src/db/group.rs
```diff
@@ -24,7 +24,7 @@ pub struct Group {
     pub icon_id: Option<usize>,
 
     /// UUID for a custom group icon
-    pub custom_icon_uuid: Option<String>,
```
louib, line 27: @sseemayer we could consider removing the `_uuid` suffix, now that the struct member is correctly typed.

src/db/meta.rs
```diff
@@ -50,22 +51,22 @@ pub struct Meta {
     pub recyclebin_enabled: Option<bool>,
 
     /// A UUID for the recycle bin group
-    pub recyclebin_uuid: Option<String>,
+    pub recyclebin_uuid: Option<Uuid>,
```
louib, line 54: @sseemayer I'm tempted to remove the `_uuid` here as well, but I feel like just `recyclebin` might be confusing :thinking: . Either way, I think we should either add the `_uuid` suffix everywhere or not use it at all. Maybe a compromise would be to use `_id` instead?

louib (approved):
What is the consequence of those changes on databases previously created by the library? I'm assuming parsing the old databases will raise an error? We could mention that in the breaking changes. Might be worth it to add a test to make sure that the database won't be loaded if we run into an invalid uuid, otherwise we might cause data corruption.

src/db/group.rs
sseemayer: I've been trying to replicate the internal XML document's structure and naming as much as possible, but agree that there might be a chance to improve ergonomics for the users of the crate by doing some gentle renaming and potentially even restructuring.

src/db/meta.rs
sseemayer: See the [other comment](https://github.com/sseemayer/keepass-rs/pull/133#discussion_r1103647333). Probably a good idea, but we should then probably rethink the whole public interface and come up with a consistent structure and naming scheme.

PR #135, opened 2023-02-11
feat!: parse colors

sseemayer:
Instead of storing raw hex strings, create a Color struct and implement parsing and dumping for that.

BREAKING CHANGE: Color fields are not Strings, but Color values now.

### DIFF
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use secstr::SecStr;
 use uuid::Uuid;
 
-use crate::db::{CustomData, Times};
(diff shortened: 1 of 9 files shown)


sseemayer:
Curiously, there doesn't seem to be a standard crate for representing colors. All libraries that I browsed through seem to be rolling their own.

louib (approved):

louib:
> Curiously, there doesn't seem to be a standard crate for representing colors. All libraries that I browsed through seem to be rolling their own.

Good thing that the implementation is straightforward! I don't mind having one less dependency for the project.

PR #139, opened 2023-02-11
docs: test all the readme examples

louib:
I also removed the `anyhow` dependency in the examples to make them easier to use, and to make the error module more visible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,18 +17,19 @@ Rust KeePass database file parser for KDB, KDBX3 and KDBX4, with experimental su
 </summary>
 
 ```rust
-use anyhow::Result;
 use keepass::{
     db::NodeRef,
-    Database, DatabaseKey
+    Database,
+    DatabaseKey,
+    error::DatabaseOpenError
 };
 use std::fs::File;
 
-fn main() -> Result<()> {
+fn main() -> Result<(), DatabaseOpenError> {
     // Open KeePass database
     let path = std::path::Path::new("tests/resources/test_db_with_password.kdbx");
     let db = Database::open(
-        &mut File::open(path)?,         		// the database
+        &mut File::open(path)?,                 // the database
         DatabaseKey::with_password("demopass"), // password (keyfile is also supported)
     )?;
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,16 +64,15 @@ fn main() -> Result<()> {
...
 use keepass::{
     db::{Database, Entry, Group, Node, NodeRef, Value},
-    DatabaseKey
+    DatabaseKey,
 };
 use std::fs::File;
 
-fn main() -> Result<()> {
-    let mut db = Database::new(Default::default())?;
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
sseemayer, line 74: @louib oh nice, thanks for showing me!

README.md
```diff
@@ -63,16 +64,15 @@ fn main() -> Result<()> {
 
 You can enable the experimental support for saving KDBX4 databases using the `save_kdbx4` feature.
 
-```rust ignore
-use anyhow::Result;
+```rust
```
sseemayer, line 67: @louib why I had previously ignored this test is that otherwise, the test suite will fail if we don't run with `--all-features`.

sseemayer (commented):
I like the general direction, but there is still an issue with this as the doc tests will fail without `--all-features` enabled (currently, they fail because of missing the anyhow crate).

Do you have an idea on how we could run the database saving example only when that feature is enabled?

louib:
> Do you have an idea on how we could run the database saving example only when that feature is enabled?

Have a look at the latest commit. The examples in the README won't be tested by default, but I think it's a good enough compromise since we will test them in CI when opening a PR.

README.md
louib: I addressed that in the latest commit.

sseemayer:
@louib I think I have come up with an even better solution - we always test the example code about database creation, but put the part that is calling the saving behind a feature check.

sseemayer (approved):

louib:
@sseemayer I think there's a potential drawback with the last commit: If users copy the example from the README, they will need to create a `save_kdbx4` feature in **their** crate, or they will have to remove the `cfg` line. This make the example more difficult to use.

sseemayer:
You are right about that. We could maybe add a comment to clarify that users should remove the line and instead ask for the feature in their Cargo.toml

PR #140, opened 2023-02-11
chore: release 0.6.0-rc.1

sseemayer:


sseemayer:
@DavidVentura @leaty @Narigo @sts10 Since you are maintainers of projects that depend on keepass-rs, I would appreciate you having a look at this release candidate. I would appreciate any pointers on ergonomics issues that you have encountered in using the crate, or ideas on how your life could be made easier!

sts10:
I feel honored to be tagged! 

One issue I think I'm seeing is that this RC can't open a KeePass database that is protected by BOTH a password _and_ a keyfile. Am I reading it right? 

I think having both is a decently common setup for users (though the password may be trivial, like "password"). 

in v0.5.X, we could just [pass keyfile as an Option, along with a password](https://github.com/sts10/medic/blob/main/src/entries.rs#L46-L50): 
```rust
  Database::open(
        &mut db_file,                                 // the database
        Some(&db_pass),                               // password
        keyfile.as_mut().map(|f| f as &mut dyn Read), // keyfile
    )
```

I see that this RC uses the new `DatabaseKey`, but seems to only accept either `with_password` OR `with_keyfile`?

sseemayer:
> I feel honored to be tagged!

I feel honored that you are using `keepass-rs` , and thanks so much for your feedback 🚀 !

> I see that this RC uses the new `DatabaseKey`, but seems to only accept either `with_password` OR `with_keyfile`?

There is also [`DatabaseKey::with_password_and_keyfile`](https://github.com/sseemayer/keepass-rs/blob/v0.6.0-rc.1/src/key.rs#L81-L86) or you can also construct `DatabaseKey` yourself:

```rust
Database::open(
    &mut db_file,
    DatabaseKey {
        password: Some(&db_pass), 
        keyfile: keyfile.as_mut().map(|f| f as &mut dyn Read),
    }
)
```

Looking at the design again, I am not 100% if it really is an improvement, so I have opened up a discussion in #143 to potentially refine the design further and hopefully make it great.

sts10:
Ah, figured I was missing something. Using your provided code snippet, [Medic](https://github.com/sts10/medic) passes all its tests, so I'm happy. 

Beyond that, I'll try to take a look at #143 this week and see if I can add anything useful. It's been a while since I looked at this part of Medic!

PR #147, opened 2023-02-12
ci: create dedicated merge workflow

louib:
This separates the PR workflow from the on-merge workflow. The `merge` workflow will also be used for automated releases in the future. We run a less extensive set of tests when merging, because we run most of the checks on the PRs. @sseemayer I would suggest enabling the `Require branches to be up to date before merging` option, to make sure that the PRs are tested against the latest version of the `master` branch.


louib:
@sseemayer We also need to update the required status checks because the commit linting job was renamed. I suggest making the code coverage job a required status check as well.

sseemayer (approved):

PR #149, opened 2023-02-12
feat: support entry updates

louib:
This PR adds an `Entry` function to commit changes to the history and to update the modification timestamp. This new function also encapsulates some expectations about the history, like the fact that recursive entry histories should not be allowed.

### DIFF
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use chrono::NaiveDateTime;
 use secstr::SecStr;
 use uuid::Uuid;
 
@@ -34,8 +35,13 @@ pub struct Entry {
 }
 impl Entry {
     pub fn new() -> Entry {
+        let mut times = Times::default();
+        let now = Times::now();
+        times.set_creation(now);
+        times.set_last_modification(now);
         Entry {
             uuid: Uuid::new_v4(),
+            times,
             ..Default::default()
         }
     }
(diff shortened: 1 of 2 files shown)


sseemayer:
Instead of modifying Entry in place, have you thought about `begin_update` returning some kind of a handle to the Entry that will accept modifications, and only change the entry (updating the history) once a `commit` method is called on the handle?

louib:
@sseemayer I love the idea! I'm going to wait for https://github.com/sseemayer/keepass-rs/pull/150 to be merged and work on that.

louib:
@sseemayer I ended up going with yet another approach: I use the history to determine if the entry was modified since the last commit. This removes the need to add a new entry field, and reduces the number of new public `Entry` functions to one. However, the implementation is not as efficient as it could be, because I have to clone the entries to sanitize them (remove the modification timestamp, for example) before the comparison. Let me know what you think!

src/db/entry.rs
```diff
@@ -110,6 +116,60 @@ impl<'a> Entry {
...
+
+    /// Adds the current version of the entry to the entry's history
+    /// and updates the last modification timestamp.
+    /// The history will only be updated if the entry has
+    /// uncommited changes.
+    ///
+    /// Returns whether or not a new history entry was added.
+    pub fn update_history(&mut self) -> bool {
+        if self.history.is_none() {
+            self.history = Some(History::default());
```
sseemayer, line 128: Since we have entry tags also being children of the history structs as well, if a user tries to edit a history entry, they would create a history which is valid in our library but KeePassXC and others will refuse to open the database.

sseemayer (approved):
Nice job! I like this implementation even if it is not maximally efficient. I think we can merge as-is and follow up the issue with histories inside of history entries and history enabledness / history size checks in a future PR.

src/db/entry.rs
sseemayer: A possible solution might be to disallow write access to the history by making it private and only providing read methods. This goes against the rest of the API being as public as it is, which is probably another discussion to have before 1.0

louib: @sseemayer

> A possible solution might be to disallow write access to the history by making it private and only providing read methods

I agree with that. The `entries` could be private, and `add_entry` would be the only method that can modify the state of the history.

louib: Addressed in https://github.com/sseemayer/keepass-rs/pull/151

Issue #153, opened 2023-03-11
Warnings about conflicting changes raised when merging with keepassxc-cli


louib:
I'm getting a bunch of warnings when merging databases with `keepassxc-cli merge` after using the `update_history` function.
The warnings look like this:
```
Entry of entry_title[25ba7e55577841ba8f3ca23a96d61bc4] contains conflicting changes - conflict resolution may lose data!
```
The warning is raised [here](https://github.com/keepassxreboot/keepassxc/blob/988bb9c1cdd0cdf76116c4c0caf89398342f3904/src/core/Merger.cpp#L452). I think the problems stems from the way the entries are compared [here](https://github.com/keepassxreboot/keepassxc/blob/988bb9c1cdd0cdf76116c4c0caf89398342f3904/src/core/Entry.cpp#L860)

louib:
I think the issue comes from how the `TimeInfo` field is parsed in KeePassXC [here](https://github.com/keepassxreboot/keepassxc/blob/988bb9c1cdd0cdf76116c4c0caf89398342f3904/src/format/KdbxXmlReader.cpp#L979). Creating a new `TimeInfo` variable will initialize all the default times to the current timestamp, which will be detected as drift between history items, since the different versions of an entry will probably not be parsed at the same time. This fits my experience because sometimes I don't get the warnings above, which can happen if all the history items are parsed during the same second.

PR #154, opened 2023-03-11
fix: set all default times when creating a node

louib:
Other KDBX parsers (like the one in KeePassXC)
store all the default times as fields of a class,
instead of items in a map. The result is that
default times that are not populated will be
given a default value at parsing time. This can
result in drift between an entry and its history items,
as described in https://github.com/sseemayer/keepass-rs/issues/153.
Populating all the default times is thus safer
for interoperability between KDBX4 parsers.

### DIFF
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -35,13 +35,9 @@ pub struct Entry {
 }
 impl Entry {
     pub fn new() -> Entry {
-        let mut times = Times::default();
-        let now = Times::now();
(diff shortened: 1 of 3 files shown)


varjolintu (approved):

louib:
@sseemayer for `Entry` and `Group`, we use `new` when we want to pre-populate some fields, and `default` when we want a "plain" instance. I thought it made sense to stay consistent with what `Entry` and `Group` are doing.

PR #155, opened 2023-03-15
feat: add merge feature 

louib:
This is a very early version of a merge feature. KeePassXC defines 5 merge modes, but I don't think we need all the 5 modes in the first version of this feature, if ever

## TODOs
- [x] handle location update
- [ ] handle deleted objects
- [ ] handle group updates
- [ ] move the feature behind a feature flags

### DIFF
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -1,9 +1,11 @@
 use std::collections::HashMap;
+use std::{thread, time};
 
 use chrono::NaiveDateTime;
 use secstr::SecStr;
 use uuid::Uuid;
 
+use crate::db::group::MergeLog;
 use crate::db::{Color, CustomData, Times};
 
 #[cfg(feature = "totp")]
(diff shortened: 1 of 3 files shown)


src/db/entry.rs
```diff
@@ -41,6 +41,63 @@ impl Entry {
             ..Default::default()
         }
     }
+
+    pub(crate) fn merge(&mut self, other: &Entry) {
+        if self.times.get_last_modification().unwrap()
```
droidmonkey, line 46: This is so verbose just to get a comparable time. Can we introduce a function in times to compare one times to another and you pass in an enum to choose which field (creation, modification, access)?

louib: We end up having to do multiple comparisons with the last modification times, so I now extract them early in the merge function [here](https://github.com/sseemayer/keepass-rs/pull/155/files#diff-83551f8ef922bfe6845bfd0c8ca2fab09b6be786474426c39f6c1ea3c7e37402R48). I'll keep youfr proposition in mind in case we have to compare the last modification times in other functions in the future.

src/db/group.rs
```diff
@@ -148,6 +157,156 @@ impl Group {
...
+                    }
+                }
+                Node::Entry(e) => {
+                    if e.uuid == id {
+                        return Some(e);
+                    }
+                }
+            }
+        }
+        None
```
phoerious, line 221: You should probably implement a recursive iterator, so you can do this:

```suggestion
        self.iter_recursive()
            .flat_map(|c| {
                match c {
                    Node::Entry(e) => Some(e),
                    _ => None
                }
            })
            .find(|e| e.uuid == id)
```

src/db/group.rs
```diff
@@ -148,6 +157,156 @@ impl Group {
...
+                    if e.uuid == entry.uuid {
+                        *e = entry.clone();
+                    }
+                }
+            }
+        }
+    }
+
+    pub fn find_entry_by_uuid(&self, id: Uuid) -> Option<&Entry> {
+        for node in &self.children {
```
phoerious, line 207:
```suggestion
        self.children.iter_mut().for_each(|node| {
```

src/db/group.rs
```diff
@@ -148,6 +157,156 @@ impl Group {
...
+        // TODO update locations
+        // TODO handle deleted objects
+    }
+
+    // Recursively get all the entries in the group, along with their
+    // location.
+    pub(crate) fn get_all_entries(
+        &self,
+        current_location: &NodeLocation,
+    ) -> Vec<(&Entry, NodeLocation)> {
```
phoerious, line 289: This should be replaced with a recursive iterator implementation as suggested above.

phoerious (changes requested):
I haven't checked the correctness of what this is doing yet, but I think the code can be streamlined quite a lot.

phoerious:
I read through the code a bit more. From what I understood, the `NodeIter` iterator is already recursive, but the traversal is breadth-first. By using a `Vec<NodeRef<'a>>` instead of a `VecDeque<NodeRef<'a'>>` as the internal traversal data structure, this could be changed to depth-first easily.

src/db/group.rs
louib: Not sure what's the advantage of this proposition. We don't need a mutable reference to the node, and the proposed syntax is more verbose.

phoerious: The mut was a bit of an oversight. iter() should be sufficient. Overall, the iterator syntax should be preferred. It's cleaner and often faster than the for-loop.

louib: @phoerious 

> often faster than the for-loop.

Do you have a source for that? I would like to know why that's the case.

> It's cleaner 

I personally prefer the for syntax, especially in Rust. Notice the absence of parentheses on the line where the `for` is declared :sparkles:

ssrlive:
Please add `From` trait to `Node` struct.
```rust
impl From<Entry> for Node {
    fn from(e: Entry) -> Self {
        Node::Entry(e)
    }
}

impl From<Group> for Node {
    fn from(g: Group) -> Self {
        Node::Group(g)
    }
}
```
and add a member function `is_group` for it:
```
    pub fn is_group(&self) -> bool {
        match self {
            Node::Group(_) => true,
            Node::Entry(_) => false,
        }
    }
```

louib:
Superseded by https://github.com/sseemayer/keepass-rs/pull/201

PR #158, opened 2023-03-24
fix: CustomData changes

varjolintu:
The CustomData in KeePassXC is handled as a hashmap, and it should processed in a similar way. Without the hashmap own iterators, setters and getters should be implemented instead of just using `.contains_key()` etc. directly. I was tempted to remove the `items` inside the `CustomData` so it could be just a define for the `HashMap<String, CustomDataItem>` but leaving as it is enabled some custom implementations for it in the future.

CustomData is now added to the Group too.

Downside of this change is that another struct is needed when parsing the data from XML because there the `<Item>` includes also the key. When dumping the data it has to make sure everything is put inside an `<Item>` again.

Some things I noticed while writing new tests. Should the parsing/dumping also handle situations where the key doesn't exist? For example the following tests are not working nicely if some incorrect data has been entered:
```
 // For fields
parse_test_xml::<SimpleTag<String>>("<TestTag>42</TestTag>");
// For CustomData
parse_test_xml::<CustomDataItemFromXml>("<Item><Value>NoKey</Value><LastModificationTime>1234</LastModificationTime></Item>");
```

Any suggestions are welcome.

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -235,16 +235,23 @@ impl Times {
 #[derive(Debug, Default, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct CustomData {
-    pub items: Vec<CustomDataItem>,
+    pub items: HashMap<String, CustomDataItem>,
 }
 
-/// Custom data field for an entry or metadata
+/// Custom data field for an entry or metadata for internal use
 #[derive(Debug, Default, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "serialization", derive(serde::Serialize))]
 pub struct CustomDataItem {
-    pub key: String,
     pub value: Option<Value>,
     pub last_modification_time: Option<NaiveDateTime>,
+}
+
+/// Custom data field for an entry or metadata from XML data
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serialization", derive(serde::Serialize))]
+pub struct CustomDataItemDenormalized {
+    pub key: String,
+    pub custom_data_item: CustomDataItem,
 }
 
 /// Binary attachments stored in a database inner header
--- a/src/xml_db/dump/mod.rs
+++ b/src/xml_db/dump/mod.rs
@@ -203,8 +203,13 @@ impl DumpXml for CustomData {
     ) -> Result<(), xml::writer::Error> {
         writer.write(WriterEvent::start_element("CustomData"))?;
 
-        for item in &self.items {
+        for (key, item) in &self.items {
+            writer.write(WriterEvent::start_element("Item"))?;
+
+            SimpleTag("Key", key).dump_xml(writer, inner_cipher)?;
             item.dump_xml(writer, inner_cipher)?;
+
+            writer.write(WriterEvent::end_element())?;
         }
 
         writer.write(WriterEvent::end_element())?;
@@ -219,10 +224,6 @@ impl DumpXml for CustomDataItem {
         writer: &mut EventWriter<E>,
         inner_cipher: &mut dyn Cipher,
     ) -> Result<(), xml::writer::Error> {
-        writer.write(WriterEvent::start_element("Item"))?;
-
-        SimpleTag("Key", &self.key).dump_xml(writer, inner_cipher)?;
-
         if let Some(ref value) = self.value {
             value.dump_xml(writer, inner_cipher)?;
         }
@@ -231,7 +232,6 @@ impl DumpXml for CustomDataItem {
             SimpleTag("LastModificationTime", value).dump_xml(writer, inner_cipher)?;
         }
 
-        writer.write(WriterEvent::end_element())?;
         Ok(())
     }
 }
--- a/src/xml_db/parse/mod.rs
+++ b/src/xml_db/parse/mod.rs
@@ -12,7 +12,8 @@ use xml::{name::OwnedName, reader::XmlEvent, EventReader};
 use crate::{
     crypt::ciphers::Cipher,
     db::{
-        Color, CustomData, CustomDataItem, DeletedObject, DeletedObjects, Group, Meta, Times, Value,
+        Color, CustomData, CustomDataItem, CustomDataItemDenormalized, DeletedObject,
+        DeletedObjects, Group, Meta, Times, Value,
     },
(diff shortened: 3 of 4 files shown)


src/xml_db/dump/mod.rs
```diff
@@ -203,8 +203,13 @@ impl DumpXml for CustomData {
     ) -> Result<(), xml::writer::Error> {
         writer.write(WriterEvent::start_element("CustomData"))?;
 
-        for item in &self.items {
+        for (key, item) in &self.items {
+            writer.write(WriterEvent::start_element("Item"))?;
+
+            SimpleTag("Key", key).dump_xml(writer, inner_cipher)?;
```
louib, line 209: @varjolintu how about we leave the key in the `CustomDataItem`, even if we index the items in the `CustomData` struct? I feel like it would keep us closer to the data format, and would still provide for easier access of the items. It would also avoid the need to update the dumping code.

src/xml_db/parse/mod.rs
```diff
@@ -511,8 +512,16 @@ impl FromXml for CustomData {
             match event {
                 SimpleXmlEvent::Start(name, _) => match &name[..] {
                     "Item" => {
-                        let item = CustomDataItem::from_xml(iterator, inner_cipher)?;
-                        out.items.push(item);
+                        let item = CustomDataItemFromXml::from_xml(iterator, inner_cipher)?;
```
louib, line 515: @varjolintu if we preserved the `key` field in the `CustomDataItem`, we could simply add a function to `CustomData` that would take care of adding the item to the HashMap. Something as simple as `CustomDataItem::add_item` would enable us to keep the parsing and dumping code unchanged. To only problem that I see is that we might end up with inconsistencies between the key of the hashmap and the key of the `CustomDataItem`, but we could make the `items` field private and provide a few public functions to read and update the items. What do you think?

src/db/group.rs
```diff
@@ -32,6 +32,9 @@ pub struct Group {
     /// The list of time fields for this group
     pub times: Times,
 
+    // Custom Data
+    pub custom_data: CustomData,
```
louib, line 36: @varjolintu I think this would have been a good candidate for a separate PR. Adding the custom data to the group struct looks like a self-contained change that wasn't necessarily related to the other changes in this PR. Moving that to a separate PR would have made reviewing this PR easier.

louib (commented):
@varjolintu I feel like this is going to improve the usability of the library!

The PR contains breaking changes, and the commits should indicate that. Have a look at the [conventional commit doc](https://www.conventionalcommits.org/en/v1.0.0/) regarding the way to do this. Also, all the commits in the PR have to follow the conventional commit convention.

src/xml_db/dump/mod.rs
varjolintu: It would keep us closer to the database format, but I think it would decrease the usability. More comments below in the other reply.

src/xml_db/parse/mod.rs
varjolintu: This was actually the whole reason of the PR. If a `HashMap` is not used we must add our own implementation of `add_item`, `remove_item`, `contains` etc. instead of just using what `HashMap` already provides. Sure, we could add some custom implementation under `CustomData`. After this change the usability is closer to how KeePassXC does things. There `CustomDataItem` only contains the value and last modification time.

Even if the database format has a separate variables for key and value, it doesn't mean it should be used directly that way. But sure, it could be a compromise that we keep the `key` in the struct also. It just doesn't have any real meaning, and that means we must add our own `add_item()` because users might forgot to set it.

src/db/group.rs
varjolintu: I agree.

src/xml_db/parse/mod.rs
varjolintu: In short, we have a few options:
- Keep the database format intact and all clients using the library must make their own implementations how to handle CustomData.
- We make `items` private and implement our own functions for handling the data.
- Use `HashMap` as the public `items` which allows using it directly, or adding more implementation around it.
- Use `items` as a private `HashMap`, but provide a public function that returns it.

src/db/mod.rs
```diff
@@ -235,18 +235,25 @@ impl Times {
...
 pub struct CustomDataItem {
-    pub key: String,
     pub value: Option<Value>,
     pub last_modification_time: Option<NaiveDateTime>,
 }
 
+/// Custom data field for an entry or metadata from XML data
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serialization", derive(serde::Serialize))]
+pub struct CustomDataItemFromXml {
```
louib, line 252: @varjolintu how about `CustomDataItemDenormalized`?

src/db/group.rs
```diff
@@ -4,7 +4,7 @@ use uuid::Uuid;
 
 use crate::db::{
     node::{Node, NodeIter, NodeRef, NodeRefMut},
-    Times,
+    CustomData, Times,
```
louib, line 7: @varjolintu I think this is a leftover from adding `CustomData` to the group struct. I'm guessing it will go away after a rebase?

src/xml_db/parse/mod.rs
louib: I think you're right. The approach you selected is probably the easiest to work with. That way we get both easy access using a key and we can ensure consistency between the key and the custom item.

louib (approved):

louib (commented):
@varjolintu I had initially approved this PR, but realized after that the breaking changes are still not highlighted in the commit message.

src/db/mod.rs
varjolintu: Yeah the "From" word is not something I'd like to see in a serialization struct. Let's change that.

src/xml_db/parse/mod.rs
```diff
@@ -639,7 +648,8 @@ mod parse_test {
         config::InnerCipherConfig,
         crypt::ciphers::PlainCipher,
         db::{
-            AutoType, AutoTypeAssociation, CustomData, CustomDataItem, Entry, History, Times, Value,
+            AutoType, AutoTypeAssociation, CustomData, CustomDataItem, CustomDataItemDenormalized,
```
louib, line 651: @varjolintu `CustomDataItem` is unused here, you could remove it in another PR.

louib (approved):

PR #162, opened 2023-04-09
Add support for Argon2id

tlercher:
Rather unintrusive change to add support for Argon2id.

I guess it could be done a bit nicer if the `try_from`-Method is reworked?

Anyway, i only tested reading properly and all the Databases were created by KeepassXC.


src/config.rs
```diff
@@ -192,12 +192,24 @@ const KDF_ROUNDS: &str = "R";
 pub enum KdfConfig {
     /// Derive keys with repeated AES encryption
     Aes { rounds: u64 },
-    /// Derive keys with Argon2
+    /// Derive keys with Argon2d
     Argon2 {
```
louib, line 196: If this is actually the `2d` variant, should we not rename the struct as well?
Doesn't have to happen in this PR, it could happen before `1.x` since the API is changing quite a lot at the moment.

louib (commented):
Thanks for the PR :pray: 

There are a few CI failures (commit message linter and code formatting), please have a look :eyes:

Please add the new kdf option to the [unit test config matrix](https://github.com/sseemayer/keepass-rs/blob/b1aa7fe3b403424e5ccf974139bd45b51625726f/src/format/kdbx4/mod.rs#L112), and then this PR is ready to go :rocket:

tlercher:
> Thanks for the PR 🙏
> 
> There are a few CI failures (commit message linter and code formatting), please have a look 👀
> 
> Please add the new kdf option to the [unit test config matrix](https://github.com/sseemayer/keepass-rs/blob/b1aa7fe3b403424e5ccf974139bd45b51625726f/src/format/kdbx4/mod.rs#L112)
Done, i also added more tests to `file_read_tests.rs`.

src/config.rs
tlercher: I definitely can do it. 

If we break the API, should i add support for Aigon2i as well? (In two seperate PRs?)

louib: > If we break the API, should i add support for Aigon2i as well?

If it's offered in applications that use KDBX, then I feel like we will get the request for it at some point. Might as well break the API before we reach `1.0`.

> In two seperate PRs?

Yes please, this one is ready to merge.

louib: @tlercher FYI created https://github.com/sseemayer/keepass-rs/issues/166 to track the `Argon2i` feature work.

PR #167, opened 2023-05-14
Update chrono dependency due to RUSTSEC-2020-0071

simonsan:
https://rustsec.org/advisories/RUSTSEC-2020-0071

Workarounds

A possible workaround for crates affected through the transitive dependency in chrono, is to avoid using the default oldtime feature dependency of the chrono crate by disabling its default-features and manually specifying the required features instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,11 @@ xml-rs = "0.8"
 base64 = "0.21"
 hex-literal = "0.4"
 secstr = "0.5"
-chrono = "0.4.23"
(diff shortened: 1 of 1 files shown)


louib (commented):
Thanks for the PR!

We use the [conventional commit](https://github.com/conventional-changelog/commitlint/#what-is-commitlint) convention for the commit messages in this repo, which is why the `Lint Commit Messages` check is failing.

louib:
@simonsan `commitlint` is still complaining about the commit title, I think because of the uppercase `U`.

PR #178, opened 2023-08-13
ci: automate the releases

### LINKED ISSUE #77: Automate releases
I want to add a GHA workflow that automates releases to `crates.io`. I propose using a variant of [semantic-release](https://github.com/semantic-release/semantic-release), which uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to determine the next release number and generate the release notes. I would also add a `CI` step that validates the format of the commit message, which would make the git history more consistent and helpful. @sseemayer let me know what you think.

louib:
This PR adds two steps to the on-merge workflow that will release a new Cargo version and push a new git tag on every new `master` commit. 

The `enable_prerelease_mode` option will make sure that version `1.0.0` is not published until we consider that the library's API is stable.

@sseemayer it looks like you were already pushing tags when publishing to Cargo so we won't have anything to do on that side before merging this PR. However, you will need to populate the repo's secrets with the `CARGO_REGISTRY_TOKEN`, since I am not an owner of that crate. The token only needs the `publish-update` scope, and I suggest that you scope the token so that it can only publish the `keepass` crate.

I've tested those two steps on a library that I maintain ([aegis-vault-rs](https://github.com/louib/aegis-vault-rs)), including making sure that breaking changes won't trigger a version `1.0.0` while in prerelease mode.

Fixes #77

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -35,3 +35,74 @@ jobs:
         with:
           fail_ci_if_error: true
           verbose: true
+
+  get-next-version:
+    name: Get the next version
+    runs-on: ubuntu-latest
+    outputs:
+      version : ${{ steps.semantic-version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: paulhatch/semantic-version@976ff820fcdca81ffc95e385a07b7eff05ddccc3
+        id: semantic-version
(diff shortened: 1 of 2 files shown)


sseemayer (approved):

sseemayer:
Thanks so much for this, and sorry for my disappearance. I have set up the token from my crates.io account and also invited you to be an owner of the package on crates.io

PR #180, opened 2023-08-28
Add tests for empty root name and empty group names

Narigo:
Hi @sseemayer,

thanks for the heads up a long while ago that you've updated the library - I finally tried updating to 0.6.1 😅

While upgrading, I added some additional tests. It seems like the current state does not allow empty group names to be used inside the kdbx files.

To check this, I have created two kdbx4 databases with empty group names. One has the root as empty folder and another one with two empty group names below the `Root` folder.

I suspect the length of the children should be two in the second test, but you may want to check this again.

The password for the kdbx files is `demopass`.


louib:
@Narigo thanks for the PR!

> It seems like the current state does not allow empty group names to be used inside the kdbx files.

I'm not entirely convinced we should support empty group names to be honest. It feels like it could cause other problems downstream, like when displaying the groups in a UI. [This KDBX spec](https://github.com/keepassxreboot/keepassxc-specs/blob/master/kdbx-xml/rfc.md#groups) doesn't tell us whether we should support empty group names or not. What is your use-case for empty group names? How did you discover this behavior?


We could also improve the error that we throw. Currently we raise the following error:
```
DatabaseIntegrity(Xml(BadEvent { expected: "text containing a value", event: End("Name") }))
```
We could explicitly handle the case and throw a specific error for empty group names. This would improve the UX of the library in the short term and form the basis of the code to handle empty group names as well.

Narigo:
I could create these empty group names with KeepassXC on Mac and opening them through the library failed for me.

In the spec you've linked, there is this:

> In addition, a group can have any of the following child elements at most once:
>
> `<Name>` : A user-specified name for this group.

I read this as "could be there, but doesn't have to be there". So throwing an error seems wrong to me.

For my use case / DX, it would be best to just receive an empty string for the name, but I do understand that a library may want to give the user a hint that there was no name being set at all vs an empty string value.

I guess the best way to deal with it would be to make the `name` an `Option<string>`, so everybody can decide on what they do with it and how to display it. What do you think?

louib:
> I guess the best way to deal with it would be to make the name an Option<string>, so everybody can decide on what they do with it and how to display it. What do you think?

I'm not opposed to changing `name` to `Option<string>`. The library hasn't reached 1.0 yet, so we can still roll out that change without too much communication overhead. I would review work going in that direction.

I think we could test all the new behavior by adding unit tests in `xml_db/parse`. I added a basic unit tests [here](https://github.com/Narigo/keepass-rs/pull/1). Using unit tests makes it easier to review and removes the need to describe the content of the new databases. A lot of the integration tests we have date back to before we had unit tests, and I believe some of them could be migrated to unit tests eventually. The two test cases you added could be replaced by self-explanatory lines of code in the unit testing modules. I think that would make your PR easier to review, and the test cases easier to maintain.

louib:
@Narigo The `Code Coverage` step should pass if you rebase this branch. The issue was fixed in [this PR](https://github.com/sseemayer/keepass-rs/pull/183)

sseemayer:
FWIW, I've rebased the branch. Apart from making the types `Option<String>`, we could also change the XML parser so that parsing an empty field that expects a `String` will yield an empty `String`, but I am not sure if that has any unexpected side effects.

PR #181, opened 2023-09-10
updates chrono dependency to avoid a security warning

sts10:
I spotted a security warning in Medic that I think this PR will fix. All keepass-rs tests pass.


sts10:
Whoops, it looks like `chrono` isn't yet using a safe version of `time`: https://github.com/chronotope/chrono/blob/main/Cargo.toml#L49

PR #182, opened 2023-10-13
refactor: delay the generation of key elements

louib:
This will make it easier to introduce challenge-response keys, since we can perform the challenge before the final call to `get_key_elements()`.

### DIFF
--- a/src/format/kdbx4/mod.rs
+++ b/src/format/kdbx4/mod.rs
@@ -81,15 +81,12 @@ mod kdbx4_tests {
             password += &std::char::from_u32(random_char as u32).unwrap().to_string();
         }
 
-        let key_elements = DatabaseKey::new()
-            .with_password(&password)
-            .get_key_elements()
-            .unwrap();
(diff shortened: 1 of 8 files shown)


src/format/kdbx4/mod.rs
```diff
@@ -81,15 +81,12 @@ mod kdbx4_tests {
             password += &std::char::from_u32(random_char as u32).unwrap().to_string();
         }
 
-        let key_elements = DatabaseKey::new()
-            .with_password(&password)
-            .get_key_elements()
-            .unwrap();
+        let db_key = DatabaseKey::new().with_password(&password);
```
varjolintu, line 84: A random question, but can this fail? What if empty password is entered?

louib: @varjolintu I think empty passwords are supported by the spec, and will result in different key elements than if no password is used for encryption.

varjolintu (approved):

PR #184, opened 2023-10-29
feat: allow using the UUID to search for nodes

louib:
This commit adds a few internal abstractions to make it easier to work with complex node paths. Using the UUID instead of the title or name is sometimes desirable because nothing guarantees that the entry titles or group names are unique. I also added a few `get` and `get_mut` tests to make sure there was no regression. The new types and functions are not public since I don't think they are stable just yet.

This is a prerequisite to https://github.com/sseemayer/keepass-rs/pull/155

### DIFF
--- a/src/db/node.rs
+++ b/src/db/node.rs
@@ -2,6 +2,45 @@ use std::collections::VecDeque;
 
 use crate::db::{entry::Entry, group::Group};
 
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub(crate) enum NodePathElement<'a> {
+    #[allow(dead_code)]
+    UUID(&'a str),
+    Title(&'a str),
+}
+
+pub(crate) type NodePath<'a> = Vec<NodePathElement<'a>>;
+
+impl<'a> NodePathElement<'_> {
+    pub(crate) fn matches(&self, node: &Node) -> bool {
+        let uuid = match node {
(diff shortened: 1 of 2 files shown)


varjolintu (approved):

sseemayer (approved):

louib:
@sseemayer @varjolintu I gave this one another shot and managed to have a static interface for both functions. Please have another look :eyes:

src/db/node.rs
```diff
@@ -2,6 +2,45 @@ use std::collections::VecDeque;
...
+
+impl<'a> NodePathElement<'_> {
+    pub(crate) fn matches(&self, node: &Node) -> bool {
+        let uuid = match node {
+            Node::Entry(e) => e.uuid,
+            Node::Group(g) => g.uuid,
+        };
+        let title = match node {
+            Node::Entry(e) => e.get_title(),
+            Node::Group(g) => Some(g.name.as_ref()),
```
varjolintu, line 22: There's no `get_name()` for a group?

varjolintu (approved):

src/db/node.rs
louib: Not yet, but I opened [a PR](https://github.com/sseemayer/keepass-rs/pull/189) to add it.

Issue #171, opened 2023-06-04
impl Zeroize for DatabaseKey


sseemayer:
As suggested [here](https://github.com/sseemayer/keepass-rs/pull/168#discussion_r1212567960), implementing [`Zeroize`](https://docs.rs/zeroize/latest/zeroize/) would increase the security of the key data in memory.

sseemayer:
This was implemented in #187

PR #187, opened 2023-10-31
feat: implement zeroize for DatabaseKey

louib:
I also added a utility function to make sure that the password never leaves the DatabaseKey when using a password prompt. We might consider exposing this feature from the crate at some point, but for now the `utilities` feature has to be enabled.

I think this is only a partial solution and there are probably other features from the `secrecy` crate that we might want to add in the future.

Addresses https://github.com/sseemayer/keepass-rs/issues/171

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ cbc = "0.1"
 
 uuid = { version = "1.2", features = ["v4", "serde"] }
 getrandom = { version = "0.2", features = ["std"] }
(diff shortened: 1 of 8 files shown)


varjolintu (approved):
Great! I have been waiting some secure memory features.

sseemayer (approved):
Nice! I agree about exposing the prompt via a separate feature flag, this seems like something that users of the crate might want.

Issue #165, opened 2023-04-16
Add support for YubiKey Challenge-Response


louib:
Something similar to what is done in KeePassXC, which can optionally use the HMAC-SHA1 challenge response to unlock the database.
I found the [yubikey.rs](https://github.com/iqlusioninc/yubikey.rs) library, but I'm not sure that it supports HMAC-SHA1 yet. Also, I tried to compile it real quick and I think it requires some C libraries. Ideally we would use a pure Rust implementation, but I'm not sure if there is one available.

PR #190, opened 2023-11-12
feat: add support for challenge-response keys

louib:
This commit adds the basic structure for challenge-response keys, as well as a `LocalChallenge` version of the
`ChallengeResponseKey`, which can be used to recover a database using the HMAC-SHA1 secret directly. This will allow us to replace the [keepassxc-cr-recovery](https://github.com/keepassxreboot/keepassxc/tree/a3717c7acd87d5e34439aa1f9f17db767a6470f3/utils/keepassxc-cr-recovery) utility.

Currently, the function to perform the challenge is synchronous, but I think in the future we might also want to add an async version, so that the feature can be used without blocking the clients' UIs.

I only implemented challenge-response for the kdbx4 format. Should I also implement it for kdbx3?

Addresse https://github.com/sseemayer/keepass-rs/issues/165


src/format/kdbx4/parse.rs
```diff
@@ -68,6 +68,9 @@ pub(crate) fn decrypt_kdbx4(
     // parse header
     let (outer_header, inner_header_start) = parse_outer_header(data)?;
 
+    #[cfg(feature = "challenge_response")]
+    let db_key = db_key.clone().perform_challenge(&outer_header.kdf_seed)?;
```
louib, line 72: I'm not a fan of having to clone the key here, but I don't see another solution if we want to keep passing the key as a reference.

varjolintu (approved):

PR #196, opened 2023-12-03
refactor: re-use header_data variable

louib:
The header_data variable is already defined in this function to validate the HMAC, and re-using it makes it easier to read the header checksum verification.


varjolintu (approved):

droidmonkey:
That's what I figured, does Rust not support variable args?

louib:
@droidmonkey I think the only option in Rust would be to use a macro.

PR #197, opened 2023-12-09
feat: add yubikey challenge-response

### LINKED ISSUE #165: Add support for YubiKey Challenge-Response

louib:
This commit adds support for YubiKey challenge-response keys, and also adds a few utilities so that we can test the new feature. I only tested on Linux, but would be interested to know if it works on Mac and Windows. I also added a utility to recover a database by using the HMAC-SHA1 secret directly. This will allow us to deprecate the [`keepassxc-cr-recovery`](https://github.com/keepassxreboot/keepassxc/tree/develop/utils/keepassxc-cr-recovery) tool from keepassxc @droidmonkey 

The new unit test is ignored since a properly-configured yubikey is required for the test to pass.

Fixes #165

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -5,10 +5,26 @@ use xml::name::OwnedName;
 use xml::reader::{EventReader, XmlEvent};
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
+#[cfg(feature = "challenge_response")]
+use challenge_response::{
+    config::{Config, Mode, Slot},
+    Yubico,
+};
+
 use crate::{crypt::calculate_sha256, error::DatabaseKeyError};
 
 pub type KeyElement = Vec<u8>;
 pub type KeyElements = Vec<KeyElement>;
+
+#[cfg(feature = "challenge_response")]
+fn parse_yubikey_slot(slot_number: &str) -> Result<Slot, DatabaseKeyError> {
+    if let Some(slot) = Slot::from_str(slot_number) {
+        return Ok(slot);
+    }
+    return Err(DatabaseKeyError::ChallengeResponseKeyError(
+        "Invalid slot number".to_string(),
+    ));
+}
 
 fn parse_xml_keyfile(xml: &[u8]) -> Result<KeyElement, DatabaseKeyError> {
     let parser = EventReader::new(xml);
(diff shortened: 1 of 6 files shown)


src/key.rs
```diff
@@ -78,6 +97,34 @@ impl ChallengeResponseKey {
                     crate::crypt::calculate_hmac_sha1(&[&challenge], &secret_bytes)?.to_vec();
                 Ok(response)
             }
+            ChallengeResponseKey::YubikeyChallenge(slot_number) => {
+                let mut yubikey_client = Yubico::new();
+                let slot = parse_yubikey_slot(slot_number)?;
+
+                let yubikey_device = match yubikey_client.find_yubikey() {
```
droidmonkey, line 104: This doesn't seem to find a specific yubikey, so if more then one is plugged in you'll get the first one in the USB chain.

louib: @droidmonkey that is correct, and I realize that we are probably doing the same thing in `keepassxc-cli`. I [opened a PR](https://github.com/wisespace-io/yubico-manager/pull/13) in the yubico-manager library to at least return all the yubikeys connected to the system. With that PR we can at least detect that multiple yubikeys are connected, and print a warning or return an error. Ideally we would be able to select the yubikey using the serial number, but that won't be as straightforward to add to the library.

droidmonkey: I already did that work, which is why we embedded the ykcore code into our code base. They wouldn't accept the PR I gave them before. In the cli we allow specifying yubikey by slot and optionally serial `slot[:serial]`

louib: I updated the PR to support multiple yubikeys. I added a `get_yubikey` function that takes an optional serial number. If the serial number is not provided, we assume that there is exactly one yubikey present on the system, and throw an error if that is not the case. Otherwise, we search for the yubikey with that serial number.

This PR is based on an [unmerged branch](https://github.com/wisespace-io/yubico-manager/pull/16) in the yubico-manager repository, so I'm going to convert this PR to a draft until the yubico-manager PR is merged.

louib: @droidmonkey The new feature is now using a stable release of [the `challenge-response` library](https://github.com/louib/challenge-response), so this PR is ready for another review.

sseemayer (approved):
Looks good to me, but I don't have a yubikey to test it

sseemayer (approved):

PR #200, opened 2023-12-18
chore: configure rustfmt

louib:
I've been wanting to do that for a while. The default `max_width` for `rustfmt` is 100, which results in many split lines, given the verbosity of Rust. The PR should not be too difficult to review, but you don't have to trust me. You can copy the `rustfmt.toml` file and run the following script, and you should get to the same result:
```
#!/usr/bin/env bash
files=$(find . -name '*.rs')
IFS=$'\n'; for file in $files; do
    rustfmt "$file"
done
```

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,3 @@
+# https://rust-lang.github.io/rustfmt for all the options.
+merge_derives = false
+max_width = 112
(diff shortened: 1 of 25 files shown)


sseemayer (approved):

droidmonkey:
We use 120 in our clang format, I'm all for longer lines

louib:
@droidmonkey yeah I don't mind longer lines either, we could definitely consider bumping to 120.

PR #201, opened 2023-12-29
feat: add merge feature

louib:
This is a re-implementation of the merge feature, which I initially attempted to implement in https://github.com/sseemayer/keepass-rs/pull/155

This new implementation is heavily inspired from the KeePassXC implementation.

I've been using this commit for over 6 months without any issues. I'd like to get the PR merged so that I don't have to rebase constantly, even though there are still some features missing. I prefixed the name of the cargo feature with an underscore so that it is hidden from docs.rs. I'll remove the underscore once we have feature parity with the KeePassXC implementation. The main merge feature missing is database metadata merging. I will add an issue to track this once this PR is merged.

Apart from the missing features, my main concern with the current implementation is that it is not as efficient as it could be. The main reason for this is that the children of a node are stored as a list, which makes it inefficient to find a node in the tree even if we have the path to that node. Optimizing that would require a bigger refactoring, and I honestly don't think it's worth it for a first draft of the feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ serialization = ["serde", "serde_json", "chrono/serde"]
 totp = ["totp-lite", "url", "base32"]
 save_kdbx4 = []
 challenge_response = ["sha1", "dep:challenge_response"]
+_merge = []
 
 default = []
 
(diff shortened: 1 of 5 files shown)


sseemayer (approved):

sseemayer:
This looks great already! Thanks for teaching me some new cfg flags 😀

PR #218, opened 2024-04-13
save_kdbx4: Fix write corruption for large Databases

### LINKED ISSUE #219: save_kdbx4: Chunking by block size in hmac_block_stream corrupts database
Hello,

Thanks for developing this crate. I think I've stumbled upon a bug that affects the save_kdbx4 feature. 
In `hmac_block_stream.rs`, this piece of code seems to cause Database writes to get corrupted:
```
    while pos < data.len() {
        let size = std::cmp::min(DEFAULT_BLOCK_SIZE as usize, data.len() - pos);
```
The issue can be fixed by removing the logic that uses the block size:
```
    while pos < data.len() {
        let size = data.len() - pos;
```
In #218, I've added a test that guards against the corruption and the fix.

baranyildirim:
In this PR:
- Add test that writes and reads back a large database (100000). This exposes the issue with write corruption, which is not detected by other tests.
- Fix write corruption by removing chunking behavior. 
- The tests fail on master, but pass after the chunking is removed.

Closes #219

### DIFF
--- a/src/hmac_block_stream.rs
+++ b/src/hmac_block_stream.rs
@@ -6,8 +6,6 @@ use crate::error::{BlockStreamError, CryptographyError};
 
 pub const HMAC_KEY_END: [u8; 1] = hex!("01");
 
-const DEFAULT_BLOCK_SIZE: u32 = 1024 * 1024;
-
 /// Read from a HMAC block stream into a raw buffer
 pub(crate) fn read_hmac_block_stream(
     data: &[u8],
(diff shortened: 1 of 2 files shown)


sseemayer:
Thanks a lot for this fix! 👍🏻 👍🏻  - I changed your commit messages to conform to our required format so that CI checks pass but left everything else as-is.

sseemayer (approved):

PR #223, opened 2024-04-24
ci: pause code coverage before releasing

louib:


droidmonkey (approved):

sseemayer:
Closing in favor of #224 - it should no longer be needed now that the codecov token is configured.

PR #230, opened 2024-08-20
Fix issues in windows

ssrlive:


sseemayer (approved):

src/key.rs
```diff
@@ -69,7 +69,11 @@ fn parse_xml_keyfile(xml: &[u8]) -> Result<KeyElement, DatabaseKeyError> {
     if key_version == Some("2.0".to_string()) {
         // TODO we should also validate the integrity of a v2 keyfile using the hash value
 
-        let trimmed_key = key_value.trim().replace(" ", "").replace("\n", "");
+        let trimmed_key = key_value
+            .trim()
+            .replace(" ", "")
+            .replace("\n", "")
+            .replace("\r", "");
```
louib, line 76: @ssrlive We should probably add a unit test for this.

ssrlive: image

louib: But this is a test run from your branch and your fork. We don't have a unit test for this upstream, right?

louib: @ssrlive gotcha, I thought we needed a new keyfile to test this, but looks like we can reproduce by running the tests on Windows.

PR #240, opened 2024-09-08
fix: make TOTP issuer field optional

tadic-luka:
Both per:
- google authenticator: https://github.com/google/google-authenticator/wiki/Key-Uri-Format#issuer
- yubikey wiki: https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html

`issuer` parameter in `otpauth://` URI scheme is optional
even though it's highly recommended to be present.

This parameter was actually optional in this repo initially 
(in https://github.com/sseemayer/keepass-rs/commit/0089160024add0b5e5a36af8f2373a63aa783e1e) 
but was changed to non optional. 
I'm changing it back to optional.

BREAKING CHANGE: makes `issuer` parameter Option instead of String

### DIFF
--- a/src/db/otp.rs
+++ b/src/db/otp.rs
@@ -33,7 +33,7 @@ impl std::str::FromStr for TOTPAlgorithm {
 #[derive(Debug, PartialEq, Eq, Zeroize, ZeroizeOnDrop)]
 pub struct TOTP {
     pub label: String,
-    pub issuer: String,
+    pub issuer: Option<String>,
     pub period: u64,
     pub digits: u32,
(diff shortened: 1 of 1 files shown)


sseemayer (approved):

louib (approved):

tadic-luka:
@louib  I can see that the commit message  CI action failed and I also saw that you updated the config to allow lines with > 100 chars. 


How do I re-run the workflow? Do I need to update the commit somehow?

louib:
@tadic-luka once my PR is merged, you can simply rebase your PR and the check should pass.

PR #241, opened 2024-09-08
chore: allow longer lines in commit messages

louib:
The 100 character limit in the commit body is very limiting, especially when we want to include URLs. This PR should make the situation slightly better, although I think in the future we should completely disable the line length check if a line is found to contain a URL. This would also allow us to preserve a smaller length limit for lines that do not container a URL.


droidmonkey:
I recommend getting rid of the limit altogether or make it something like 4096

droidmonkey (approved):

sseemayer (approved):

PR #252, opened 2025-04-19
feat(kdbx4): handle public custom data in outer header

sseemayer:
Databases created with KeepassXC now contain a KPXC_PUBLIC_UUID field in the public custom data outer header [1], causing parsing to fail because of an unrecognized outer header field. This change adds support for these fields.

[1] https://keepass.info/help/kb/kdbx.html

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -49,6 +49,9 @@ pub struct DatabaseConfig {
 
     /// Settings for the Key Derivation Function (KDF)
     pub kdf_config: KdfConfig,
+
+    /// Custom data of plugins/ports.
+    pub public_custom_data: Option<VariantDictionary>,
 }
 
 /// Sensible default configuration for new databases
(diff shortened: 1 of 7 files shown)


droidmonkey:
Why are you dropping it? Public custom data is part of the kdbx 4 specification.

sseemayer:
> Why are you dropping it? Public custom data is part of the kdbx 4 specification.
> 
> https://keepass.info/help/kb/kdbx.html#header

@droidmonkey  admittedly, I was too lazy to fully implement parsing and writing because I was working on another project where the `keepass` crate is a dependency, but I have now done the full implementation

droidmonkey (approved):

droidmonkey:
Thank you for your work on the rust implementation, maybe one day we will have the opportunity to make that the core of KeePassXC

PR #253, opened 2025-04-26
fix: make clippy happy

ssrlive:
Follow the prompts of the `clippy` command and make the following changes.

```bash
cargo clippy --all-features --all-targets -- -D warnings
```

### DIFF
--- a/src/db/group.rs
+++ b/src/db/group.rs
@@ -16,7 +16,7 @@ pub(crate) type NodeLocation = Vec<Uuid>;
 
 pub enum SearchField {
     #[cfg(any(test, feature = "_merge"))]
-    UUID,
+    Uuid,
     Title,
 }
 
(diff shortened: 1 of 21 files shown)


sseemayer (approved):

sseemayer:
Sweet, thanks! I learned some new patterns that probably weren't possible when some of that code was originally written :D

src/db/group.rs
```diff
@@ -16,15 +16,15 @@ pub(crate) type NodeLocation = Vec<Uuid>;
 
 pub enum SearchField {
     #[cfg(any(test, feature = "_merge"))]
-    UUID,
+    Uuid,
```
louib, line 19: This is a breaking change. I see that the PR was already merged, but it should probably have been pointed out in the commit description.

louib (approved):

src/db/group.rs
louib: I created https://github.com/sseemayer/keepass-rs/pull/254 since I don't think this field should even be used outside of this create.

PR #256, opened 2025-04-28
feat: set_unprotected_field_pair & set_protected_field_pair .etc

ssrlive:
It is not clear what function the "otp" key should call, please review and let me know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,8 +79,6 @@ totp-lite = { version = "2.0", optional = true }
 url = { version = "2.2", optional = true }
 base32 = { version = "0.5", optional = true }
 
-[dev-dependencies]
-rustfmt = "0.10"
 
 [[bin]]
 # parse a KeePass database and output as a JSON document
--- a/src/db/entry.rs
+++ b/src/db/entry.rs
@@ -97,6 +97,28 @@ impl Entry {
         }
     }
 
+    #[rustfmt::skip]
+    const EXCLUDED_FIELDS: [&str; 9] = ["Password", "BinaryData", "otp", "Title", "URL", "UserName", "Notes", "Additional", "BinaryDesc"];
+
+    pub fn set_additional_attribute(
+        &mut self,
+        key: &str,
+        value: Option<&str>,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        if Self::EXCLUDED_FIELDS.contains(&key) {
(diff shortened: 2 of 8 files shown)


sseemayer:
I agree with adding setters for the common fields of an entry, but would prefer not hiding the implementation of the fields behind the pub(crate) accessors.

ssrlive:
It is unsafe to allow end users to directly manipulate fields. Therefore it is appropriate to hide it using pub(crate) accessors.
If you wish, you could add another function such as set_additional_attribute to give the user the ability to add arbitrary data.
```rust
    pub fn set_additional_attribute(&mut self, key: &str, value: Option<&str>) -> crate::Result<()> {
        #[rustfmt::skip]
        let excluded_fields = ["Password", "BinaryData", "otp", "Title", "URL", "UserName", "Notes", "Additional", "BinaryDesc"];
        if excluded_fields.contains(&key) {
            return Err(format!("Cannot set additional attribute for field {}", key).into());
        }
        self.set_unprotected_field_pair(key, value);
        Ok(())
    }
```

Now I just want to know, when adding the `otp` key-value pair, should I call `set_unprotected_field_pair` or `set_protected_field_pair`?

sseemayer:
Users can already modify `fields` on an entry directly as they are `pub`; the Rust type system should ensure that all assigned values are correct and thus safe.

I am proposing to remove all of the `set_*_field_pair` methods from the PR and instead implementing `set_username`, `set_password`, `set_url`, `set_otp`, ... like this:

```rust
pub fn set_username(&mut self, username: Option<&str>) {
        if let Some(v) = username {
                self.fields.insert("UserName", Value::Protected(v.into());
        } else {                
                self.fields.remove("UserName")
        }
}
```


otp are the secrets for time-based tokens, so they should be protected.

ssrlive:
This is just a small exercise in DRY principle, I don’t understand why you must delete the `set_unprotected_field_pair` function.

Additionally, you can make `fields` private if you wish.
```rust
pub(crate) fields: HashMap<String, Value>,
```

sseemayer:
I do not want to add the `set_*_field_pair` methods because they decrease readability of the code. 

I would disagree with the DRY argument because you are adding 5 utility methods to enable the implementation of 4 setters, and the amount of repeated code is 3 lines per setter, so your "DRY" solution is likely longer and more complex overall.

ssrlive:
We were unable to convince each other, so this PR will be discarded.

PR #260, opened 2025-05-11
fix(kdbx4): writing custom data into outer header

sseemayer:
PR #252 introduced a bug where custom data would get written to the
outer header after it had already ended, leading to the opening of
written databases to fail with a header hash mismatch error. this commit
corrects the order of write operations.

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -791,8 +791,13 @@ mod database_tests {
     #[cfg(feature = "save_kdbx4")]
     #[test]
     fn test_save() {
-        use crate::db::Entry;
+        use crate::{db::Entry, variant_dictionary::VariantDictionary};
         let mut db = Database::new(Default::default());
+
+        let mut public_custom_data = VariantDictionary::new();
+        public_custom_data.set("example", 42);
+
+        db.config.public_custom_data = Some(public_custom_data);
 
         db.root.add_child(Entry::new());
         db.root.add_child(Entry::new());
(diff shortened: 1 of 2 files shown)


droidmonkey:
I think we need to introduce a test that exercises and checks this code's output, or modify an existing test to include custom data into it.

sseemayer:
> I think we need to introduce a test that exercises and checks this code's output, or modify an existing test to include custom data into it.

good idea, I have done so in [ae9f9eb](https://github.com/sseemayer/keepass-rs/pull/260/commits/ae9f9ebf8efb1569f274322da8bb853e56d42929)

droidmonkey (approved):

droidmonkey:
@sseemayer wanted to also mention we should introduce a test coverage CI so we can visualize where more testing is needed. We use codecov in keepassxc which is nice, injects a report into the PR as well.

src/db/mod.rs
```diff
@@ -791,9 +791,14 @@ mod database_tests {
     #[cfg(feature = "save_kdbx4")]
     #[test]
     fn test_save() {
-        use crate::db::Entry;
+        use crate::{db::Entry, variant_dictionary::VariantDictionary};
         let mut db = Database::new(Default::default());
 
+        let mut public_custom_data = VariantDictionary::new();
+        public_custom_data.set("example", 42);
```
louib, line 798: I feel like we're only doing half the test here. Should we not also verify that the value is still there after loading the database? that would make sure not only that the database file is valid after saving with custom data, but that the custom data is still there.

sseemayer: There is an [`assert_eq!` check in the test](https://github.com/sseemayer/keepass-rs/blob/c72d53526fde5928505952d3cde6c439f8375622/src/db/mod.rs#L817) to see if the database objects before and after serialization are equal. We `#[derive(PartialEq, Eq)` all the way down to the `VariantDictionary`, so that should be strong enough of a check.

louib: :sweat_smile: thanks for pointing that ou

PR #267, opened 2025-09-13
refactor!: restructure project for better DX and UX

sseemayer:
This PR is a major change to the internal structure and API interface in preparation for a 1.0 release of the crate.

* feat: add attachment support
* feat: add custom icon support
* feat: add history tracking editing

BREAKING CHANGE: Different database types and how they are accessed. Removes the `Node`, `NodeRef`, etc concepts.
BREAKING CHANGE: Completely replaced error types to make them more granular
BREAKING CHANGE: Structure of output generated by serde::Serialize has changed


droidmonkey:
FWIW, I had copilot agent analyze and propose a solution to transactional edits to the database. It also proposed a JSON & API based method for these transactions. This was a thought project on how we could implement immutable database/group/entry sharing in KeePassXC and move all editing into a centralized place. https://github.com/droidmonkey/keepassxc/pull/6

sseemayer:
@Its-Just-Nans , thanks for the suggestion! 

I have added these lints to `Cargo.toml`: 

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tarpaulin_include)'] }
missing_docs = "warn"
rust_2018_idioms = "warn"

[lints.clippy]
indexing_slicing = "deny"
missing_panics_doc = "deny"
unwrap_used = "deny"
expect_used = "deny"
cast_lossless = "deny"
cargo = { level = "warn", priority = -1 }
```

and fixed many many lint errors. There are many cases where `expect` and `unwrap` are used in a way that cannot cause a panic, and they are all justified in comments next to `#[allow(clippy::unwrap_used)]` now.

yoyosource:
After trying out the current released version I really like this library. Though it bugged me that I was not able to connect the attachments inside of the Header to a specific Entry. Which is why I tried out the current rev of this branch and after some refactoring I was no longer able to read my KDBX4 file. The `Database::open` complained about the `IsExpanded` property not being present. It would be great to have a sensible default (most likely `false`) for this XML Field to enable parsing of KDBX4 files that dont have this Field present.

yoyosource:
Here you go:
[IsExpandedTest.kdbx.zip](https://github.com/user-attachments/files/25292662/IsExpandedTest.kdbx.zip)
The password is `TestTestTest` it only contains 1 Group and nothing else. I created this KDBX4 with Strongbox and I tried to open it with the latest commit of this branch and still got this error:
```
Custom { kind: Other, error: Parse(KDB4(Xml(Xml(Custom("missing field `IsExpanded`"))))) }
```

yoyosource:
@sseemayer Thanks for the fix. I tried the new commit locally and it works as expected now. When I find another issue I will write here.

One more things how can I convert an Entry from a history to a EntryRef so I can access the stored attachments? Or is that impossible by design/spec?

sseemayer:
> One more things how can I convert an Entry from a history to a EntryRef so I can access the stored attachments? Or is that impossible by design/spec?

@yoyosource  there is a bit of weirdness necessary because `EntryRef` must point to a "live" Entry in the database, and the History and all historical entries is contained within a live Entry. I have now added a `restore_entry` function that allows you to promote a historical version of an Entry to a full Entry, and get whatever data from there.

See [this test](https://github.com/sseemayer/keepass-rs/blob/16361b4f59447ec113cc0a6bf610cc5400ad3566/tests/entry_tests.rs#L127) for an example on how to use.

yoyosource:
@sseemayer Wouldnt it be possible to allow accessing the Attachments on an Entry instead of only on an EntryRef instead? Since the data needs to be there regardless of having an EntryRef or an Entry. And then I would access the Attachment from the Database manually.

yoyosource:
Looks good thanks. I will try it out in a bit but the test looks like what I need.

yoyosource:
No actually there is some other issue here. Somehow it sometimes loads the Attachments from the previous Entry instead of the current one. I am still trying to find out why this happens.

sseemayer:
@yoyosource thanks a lot for the report and fix!

sseemayer:
after discussing with @louib , this PR will not be merged as-is, but rather re-implemented as a series of more reviewable changes, the first one (#290) now in draft stage. The end result will likely be very similar to this branch, though.

sseemayer:
This has now been completely replaced by #291 #292 #293 #294

PR #273, opened 2025-09-14
ci: consolidate test jobs

louib:
I initially wanted to add an `armv8` test job, since GitHub offers `armv8` runners for free for open source projects. I decided to delay this change and for now only consolidate the test jobs into a test matrix since I've waited over 10 minutes for an `armv8` runner to become available :grimacing: 

I added a `Test Suite` rollup step to give us flexibility when adding and removing unit test jobs with different targets or underlying OSes. @sseemayer this means we can remove the `ARMv7 Test Suite` required status check, since this will now be handled by the `Test Suite` rollup step.

We also have a windows job that only builds the project, but I think in a future PR I'm going to add it to the test matrix so that we're also running unit tests on Windows instead of just building the library.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,8 +72,24 @@ jobs:
         run: cargo build --no-default-features
 
   test:
-    name: Test Suite
-    runs-on: ubuntu-latest
+    name: Test Suite on ${{ matrix.name }}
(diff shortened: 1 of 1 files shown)


sseemayer (approved):

sseemayer:
Thanks a lot! I merged and updated the required status checks on mobile, hope it went well 🙈

PR #281, opened 2025-11-22
Support wasm

czlol:
## Summary

This PR adds improved `wasm32` support to `keepass-rs`, making it safer and more reliable to use in WebAssembly environments (e.g. browsers) without affecting existing native targets.

## Motivation

- When running on `wasm32`, some assumptions about time and threading do not hold (no OS clock, no native threads in many environments).
- `keepass-rs` is used in web contexts where deterministic, panic‑free behavior is important, especially around Argon2 key derivation and timestamp handling.
- The goal is to provide a minimal, targeted set of changes that make the crate “just work” on `wasm32` while keeping the native behavior unchanged.

## Implementation details

- **Target-specific time source for `Times::now`**
  - Introduce a small helper, `now_timestamp()`, that is specialized per target:
    - On `wasm32`, use `js_sys::Date::now()` to obtain the current time from the JavaScript host and convert it to a Unix timestamp in seconds.
    - On non-`wasm32` targets, keep using `chrono::Utc::now().timestamp()` as before.
  - `Times::now()` and the associated timestamp helpers are updated to go through this abstraction so that all time handling works consistently on both native and `wasm32`.

- **Safe Argon2 configuration on `wasm32`**
  - For the Argon2-based KDF, configure the thread mode based on the compilation target:
    - On `wasm32`, force `argon2::ThreadMode::Sequential` to avoid panics on platforms that do not provide WebAssembly thread support.
    - On other targets, keep using `argon2::ThreadMode::Parallel` to preserve existing performance characteristics.
  - This keeps the KDF behavior functionally equivalent while avoiding runtime failures in browser environments.

### Limitations

- **Only single-threaded mode is supported on `wasm32`**

## Compatibility

- No public API or feature flags were changed.
- Native builds (`x86_64-unknown-linux-*`, `x86_64-pc-windows-*`, etc.) should behave exactly as before.
- The WebAssembly-specific behavior is only enabled when compiling for `wasm32`, and is implemented via `cfg` guards.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,9 @@ base32 = { version = "0.5", optional = true }
 [dev-dependencies]
 rustfmt = "0.10"
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+js-sys = "0.3"
+
 [[bin]]
 # parse a KeePass database and output as a JSON document
 name = "kp-dump-json"
--- a/src/crypt/kdf.rs
+++ b/src/crypt/kdf.rs
@@ -55,8 +55,15 @@ impl Kdf for Argon2Kdf {
         &self,
         composite_key: &GenericArray<u8, U32>,
     ) -> Result<GenericArray<u8, U32>, CryptographyError> {
+        // Disable Argon2 multithreading on wasm32 targets to avoid panics on platforms without thread support.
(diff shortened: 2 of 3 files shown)


sseemayer (approved):

sseemayer:
Thanks for the contribution! Is it possible to also test the wasm32 target in out CI/CD?

PR #293, opened 2026-02-19
refactor!: move from event-based xml (de)serialization to serde-based xml (de)serialization

sseemayer:
instead of event-based XML serialization and deserialization using `xml-rs`, use `quick-xml` with the `serialization` feature to (de)serialize using structs with `#[derive(Serialize, Deserialize)]`. This decreases complexity significantly and makes it much less likely for round-tripping errors to occur.

Also implements proper attachment support for both header (KDBX4) and XML (KDBX3) attachments.

BREAKING CHANGE: Attachments are now properly handled in the user-facing API.
BREAKING CHANGE: `Database::deleted_objects` has simplified API
BREAKING CHANGE: custom_data has simplified API
BREAKING CHANGE: Times has a defined set of fields instead of a HashMap of times
BREAKING CHANGE: Entry fields have new Value type

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,10 @@ jobs:
     name: Lint Commit Messages
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
-      - uses: wagoid/commitlint-github-action@v5
+      - uses: wagoid/commitlint-github-action@v6
         with:
           configFile: commitlintrc.json
 
@@ -27,7 +27,7 @@ jobs:
     name: Build Doc
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 1
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


src/db/types/attachment.rs
```diff
@@ -1,23 +1,6 @@
-/// Binary attachments stored in a database inner header
```
louib: @sseemayer can we replace that doc with something more meaningful if the previous doc no longer is?

src/db/types/value.rs
```diff
@@ -1,33 +1,112 @@
...
+#[derive(Debug)]
 pub enum Value {
-    Bytes(Vec<u8>),
-    Unprotected(String),
-    Protected(SecStr),
+    /// unprotected text data
+    String(String),
+
+    /// protected text data
+    PString(SecretBox<String>),
```
louib: Could we call this `ProtectedString` or `SecretString`?

src/format/xml_db/meta.rs
```diff
@@ -0,0 +1,862 @@
...
+    #[serde(rename = "@Protected", default, with = "cs_opt_bool")]
+    pub protected: Option<bool>,
+}
+
+impl Binary {
+    pub(crate) fn xml_to_db(
+        self,
+        inner_decryptor: &mut dyn crate::crypt::ciphers::Cipher,
+    ) -> crate::db::Attachment {
+        let mut data = base64_engine::STANDARD.decode(self.value).unwrap_or_default();
```
louib: @sseemayer all the `unwrap_or_default` in this function are making me nervous. Should we not be propagating the errors?

src/format/xml_db/mod.rs
```diff
@@ -1,307 +1,247 @@
+//! XML (de)serialization for KeePass databases.
+//!
+//! This module provides types that mirror the ones in `crate::db`, but are tailored to closely fit
+//! the XMl structure of KeePass databases for easy `#[derive(Serialize, Deserialize)]`.
```
louib: @sseemayer nitpick: weird casing for `XML` here

sseemayer: fixed :)

PR #294, opened 2026-02-21
refactor!: attachments/entries/groups/icons owned by Database

sseemayer:
Re-design how `Attachment`, `CustomIcon`,  `Entry` and `Group` are stored in a database - from nested structs to flat HashMaps stored within the Database struct, with cross-references by e.g. `EntryId` and `GroupId`, for faster traversal. 

Invariants in how the objects can be accessed ensure that the database is always in a valid state, e.g. EntryId and GroupId refer to entries that still exist in the database. Instead of e.g. raw `&Entry`, use `EntryRef`, instead of raw `&mut Entry`, use `EntryMut` accessors that ensure safe modification and access to fields.

BREAKING CHANGE: Attachment, CustomIcon, Entry and Group access completely changed


src/format/xml_db/mod.rs
```diff
@@ -73,24 +83,50 @@ impl KeePassFile {
...
             for binary in binaries.binaries {
-                attachments.push(binary.xml_to_db(inner_decryptor)?);
+                let id = crate::db::AttachmentId::next_free(&db);
+                let data = binary.xml_to_db(inner_decryptor)?;
+
+                attachments.insert(
+                    id,
+                    crate::db::Attachment {
+                        id,
+                        entries: HashSet::new(),
```
louib, line 112: @sseemayer we could maybe have a constructor that initializes the `entries` set internally since we seem to be defaulting to an empty set multiple times

louib (approved):

PR #307, opened 2026-03-25
fix: use SHA-512 for ChaCha20 inner cipher key derivation in KDBX3

### LINKED ISSUE #300: Bug when decrypting chacha20 protected values
I was attempting to decrypt the kdbx 3.1 file generated by jetbrains's DataGrip (it's how DataGrip [stores database passwords](https://www.jetbrains.com/help/idea/reference-ide-settings-password-safe.html)) 
And I noticed that the passwords themselves are garbled up (after successfully decrypting the db)

I am not sure if this is a bug with DataGrip's kdbx implementation, or if it's a bug with this project's impl, I can't find the authoritative kdbx 3.1 spec (only the 4.1 spec [here](https://keepass.info/help/kb/kdbx.html))

The curx of the issue is incorrect ChaCha20 stream cipher key derivation.

The spec states
```
ChaCha20. K should consist of 64 bytes. Compute H := SHA-512(K). The key for ChaCha20 is (H[0], ..., H[31]), and the nonce is (H[32], ..., H[43]).
```

The Logic to create the `ChaCha20Cipher` is sound (is correctly extracts the right bytes from the sha512 hash)
https://github.com/sseemayer/keepass-rs/blob/6d83a6132583c38a003075f35cf4737f23586395/src/crypt/ciphers.rs#L173-L182

`ChaCha20Cipher::new` is called with the "stream_key" through `InnerCipherConfig::get_cipher` here https://github.com/sseemayer/keepass-rs/blob/6d83a6132583c38a003075f35cf4737f23586395/src/format/kdbx3.rs#L193

The problem is that `stream_key` is calculated as the sha-256 of the `protected_stream_key` (from header field num 10), but that is only correct for the `Salsa20` cipher, for the `ChaCha20` cipher a sha-512 of the `protected_stream_key` should be used as the key/iv.

I want to emphasize, that every implementation of kdbx I tried using was unable to read this field, Keepassxc, Keepass, KeepassX none of there correctly read this data.
 

I have a fix and a reproduction case in my fork (see this commit https://github.com/mendymm/keepass-rs/commit/ef7c5b4500fa67aff4b522934ca9f20d9c07ea45)
Would you be interested in a PR?

queelius:
## Summary

Fixes #300 - KDBX 3.1 ChaCha20 inner cipher decrypts protected values incorrectly.

- **Root cause:** The KDBX3 code path pre-hashed `protected_stream_key` with SHA-256 before passing it to cipher constructors. `ChaCha20Cipher::new()` then computed SHA-512 on this *already-hashed* value, producing `SHA-512(SHA-256(key))` instead of the spec-required `SHA-512(key)`. This caused silent decryption failure for protected fields (passwords appeared as garbled data).
- **Fix:** Moved SHA-256 hashing into `Salsa20Cipher::new()` so each cipher constructor handles its own key derivation internally, matching the existing pattern in `ChaCha20Cipher::new()`. Removed the SHA-256 pre-hash from `decrypt_kdbx3()`.
- **Added test** with a KDBX 3.1 database using ChaCha20 inner cipher, verifying that protected password fields decrypt correctly.

All 58 existing tests continue to pass (no regressions for KDBX3 Salsa20, KDBX4, or KDB formats).

## Files changed

| File | Change |
|------|--------|
| `src/format/kdbx3.rs` | Remove SHA-256 pre-hash, pass raw `protected_stream_key` to cipher |
| `src/crypt/ciphers.rs` | Add SHA-256 hashing inside `Salsa20Cipher::new()` |
| `src/config.rs` | Rename parameter to `protected_stream_key` for clarity |
| `tests/entry_tests.rs` | Add KDBX3 + ChaCha20 protected fields test |
| `tests/resources/test_db_kdbx3_with_chacha20_protected_fields.kdbx` | Test database |

## Test plan

- [x] New test `kdbx3_with_chacha20_protected_fields` passes - verifies ChaCha20 protected values decrypt correctly
- [x] Existing KDBX3 tests pass (Salsa20 path not regressed)
- [x] Existing KDBX4 tests pass (both AES and ChaCha20 inner ciphers)
- [x] All 58 tests pass with `cargo test`

## References

- KeePass spec (ChaCha20 key derivation): *"Compute H := SHA-512(K). The key for ChaCha20 is (H[0], ..., H[31]), and the nonce is (H[32], ..., H[43])."*
- Reporter's analysis and reproduction: https://github.com/mendymm/keepass-rs/commit/ef7c5b4500fa67aff4b522934ca9f20d9c07ea45

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sseemayer (approved):

sseemayer:
thanks also to @mendymm for the detailed analysis!

Issue #303, opened 2026-03-17
Error: Error parsing XML inside KDBX: unknown variant `PreviousParentGroup`, expected `Group` or `Entry`


Syndim:
Thanks for building the crate! Found a parser issue after upgrading to v0.10.1:

```
Error parsing XML inside KDBX: unknown variant `PreviousParentGroup`, expected `Group` or `Entry`
```

It looks there can be property named `PreviousParentGroup` under the group like this:

```
	<IsExpanded>True</IsExpanded>
	<DefaultAutoTypeSequence/>
	<EnableAutoType>null</EnableAutoType>
	<EnableSearching>null</EnableSearching>
	<LastTopVisibleEntry>(some value)</LastTopVisibleEntry>
	<PreviousParentGroup>(some value)</PreviousParentGroup>
```

The current parser is too strict and will fail on any unknown child properties, any chance to make it less strict?

sseemayer:
The issue is due to [additional fields added in the KDBX 4.1 format](https://keepass.info/help/kb/kdbx_4.1.html). Since I am in the middle of a major refactoring of the project (see #294), I will finish that refactor and then add support for the new fields.

Long-term, failing on unknown fields is safer since we want to be sure that we are not silently dropping fields during database write operations.

PR #312, opened 2026-04-26
Serialize nullable bools and DataTransferObfuscation per KeePassXC schema

### LINKED ISSUE #311: Files written by Database::save rejected by KeePassXC 2.7+
### Reproduction

Checkout commit https://github.com/sseemayer/keepass-rs/pull/312/commits/c43537671c693903bb9545cc606f63ac68977307. Confirm that the added tests fail. #312 follows up these failing tests with modified behavior that passes the tests.

### Diagnosis

Two distinct serializer bugs on `src/format/xml_db/`:

- `cs_opt_bool::serialize` for `None` emits `serialize_none()` (which `quick-xml` writes as a self-closing element such as `<EnableSearching/>`). KeePassXC's KDBX schema treats `EnableSearching` and `EnableAutoType` as a tri-state and requires the literal string null for the unset state (e.g. `<EnableSearching>null</EnableSearching>`). Empty elements are rejected with `Invalid` `<field>` value. The deserializer already maps `"null"` to `Ok(None)`, so the fix is a one-line symmetry change.

- `AutoType::data_transfer_obfuscation` is `Option<bool>` with `cs_opt_bool` which writes `True/False`. Per the KDBX schema (and what KeePassXC writes), this field is an integer 0/1. KeePassXC rejects the bool strings with `Invalid number value`. The cleanest local fix without breaking the public type is a new `cs_opt_intbool` helper that serializes as `0/1` and accepts either form on read.

XML diff (`KeePassXC` original vs `Database::save` round-trip)

```diff
-<EnableAutoType>null</EnableAutoType>
-<EnableSearching>null</EnableSearching>
+<EnableAutoType/>
+<EnableSearching/>
...
-<DataTransferObfuscation>0</DataTransferObfuscation>
+<DataTransferObfuscation>False</DataTransferObfuscation>
```
### Related

Issue #303 covers `<PreviousParentGroup>` parse failures (independent bug, also affects KDBX 4.1 round-trip). Happy to push more updates if you'd just like it in one PR.

davidroeca:
Closes: https://github.com/sseemayer/keepass-rs/issues/311

Two small changes to src/format/xml_db/ so that files written by `Database::save` are accepted by KeePassXC 2.7+.

### Tests

`tests/keepassxc_writer_compat_tests.rs` adds three regression tests against a real KDBX 4.1 fixture (generated by KeePassXC 2.7.12). Each test asserts on the literal XML bytes produced by Database::save, not on parser behavior. The deserializer is permissive enough that a self-roundtrip would pass even on the broken code. Verified that all three tests fail without this PR's source changes.

### Not in scope (for follow-up)

- `<PreviousParentGroup>` parse failure (issue #303). Independent bug; happy to bundle if preferred.
- Casing drift on `<DefaultUserName>`, `<DefaultUserNameChanged>`, `<ProtectUserName>`. `keepass-rs` writes Username (lowercase n). KeePassXC skips these as unknown (warnings, not errors). A separate serde(rename) cleanup.
- Database::config.version is always reported as KDB4(0) after parsing regardless of source minor version. Cosmetic but means writes are always emitted as KDBX 4.0.


sseemayer (approved):

sseemayer:
Thanks for the report and fix! I have created #313 about the casing issue, everything else mentioned would make sense to fix as part of #303

PR #316, opened 2026-05-02
feat!: add support for KDBX4.1 (#303)

### LINKED ISSUE #303: Error: Error parsing XML inside KDBX: unknown variant `PreviousParentGroup`, expected `Group` or `Entry`

sseemayer:
closes #303

Add support for the new features of KDBX4.1[1] such as group tags, disabling password quality checks for entries, previous parent groups, custom icons with names and last modification times

BREAKING CHANGE: databases will be written in KDBX4.1 and emit a DatabaseSaveError::UnsupportedVersion if writing as KDBX4.0 is requested

BREAKING CHANGE: Entry::quality_check type change from Option<bool> to bool

[1] https://keepass.info/help/kb/kdbx_4.1.html


lordgreg:
waiting for the PR, since I also rely on the new KDBX Version, which includes PreviousParentGroup. With the latest version of the package, I am getting the exception:

> Failed to open KDBX file 'foo.kdbx': Error parsing XML inside KDBX: unknown variant `PreviousParentGroup`, expected `Group` or `Entry`

sergeykonkin:
Hi all, any updates on this? I also ran into an issue with the database edited in Keepassium, but related to `Tags`:

> Error parsing XML inside KDBX: unknown variant `Tags`, expected `Group` or `Entry`

Is there an ETA for when this PR might be merged?

src/db/merge.rs
```diff
@@ -636,6 +638,122 @@ fn merge_history(dest: &History, source: &History, log: &mut MergeLog) -> Result
...
+                // both times are blank - do not re-add the icon
+                continue;
+            }
+        }
+
+        dest_db.custom_icons.insert(id, source_icon.clone());
+
+        log.events.push(MergeEvent {
+            target: MergeEventTarget::Icon(id),
+            event_type: MergeEventType::Created,
```
louib, line 677: @sseemayer would be nice to have a dedicated event type since otherwise it's impossible to tell if the target ID is an entry ID, and icon ID or a group ID

src/format/xml_db/group.rs
```diff
@@ -146,6 +158,7 @@ impl Group {
             uuid: UUID(source.id().uuid()),
             name: source.name.clone(),
             notes: source.notes.clone(),
+            tags: source.tags.join(";").into(),
```
louib, line 161: `";"` should probably be a constant somewhere instead of being hardcoded here and elsewhere

tests/common/mod.rs
```diff
@@ -194,7 +194,7 @@ pub fn fast_combo() -> Combo {
 impl Combo {
     pub fn get_config(&self) -> DatabaseConfig {
         let mut cfg = DatabaseConfig::default();
-        cfg.version = DatabaseVersion::KDB4(0);
+        cfg.version = DatabaseVersion::KDB4(1);
```
louib, line 197: Should we be using `KDBX4_CURRENT_MINOR_VERSION` ?

louib (approved):

lordgreg:
Thank you! Can't wait to test this out! :) 👍🏻

src/db/merge.rs
sseemayer: we can distinguish target type using the `MergeEventTarget` enum.

src/format/xml_db/group.rs
sseemayer: spun out into #339

PR #319, opened 2026-05-02
fix(xml_db): accept KeePassXC PreviousParentGroup element

jmylchreest:
## Summary

KeePassXC writes a `<PreviousParentGroup>UUID</PreviousParentGroup>` element on `Group` and `Entry` to record the previous parent group when items are moved between groups (used for the undo/restore feature, [KeePassXC #5544](https://github.com/keepassxreboot/keepassxc/pull/5544), shipped in 2.7.x).

`keepass-rs` currently fails to deserialize any KDBX file containing this element. On `Group`, the element falls through to the `children: Vec<GroupOrEntry>` `$value` catch-all and the deserializer errors with:

```
Error parsing XML inside KDBX: unknown variant `PreviousParentGroup`, expected `Group` or `Entry`
```

This blocks reads of any KeePassXC database after the user has moved an entry between groups (which is increasingly common — the move is recorded automatically).

## Fix

Add an explicit `previous_parent_group: Option<UUID>` named field on both `Group` and `Entry` (PascalCase rename via the existing struct-level attr, decoded with the existing `cs_opt_string` helper). Now the element is consumed by a known field instead of routing to the children stream.

The value is captured but not yet plumbed through to the in-memory `Group`/`Entry` model — round-trip preservation on save would require a follow-up. For now this is a read-fix only; the field is `skip_serializing_if = \"Option::is_none\"` so it doesn't appear when re-saving databases that didn't carry it.

## Test plan

- [x] Added regression cases in `test_deserialize_group` and `test_deserialize_entry` exercising the `<PreviousParentGroup>` element.
- [x] `cargo test --lib format::xml_db` passes (24 tests, 0 failed).
- [x] Manually validated against a real KeePassXC 2.7.x kdbx file containing moved entries — opens cleanly post-fix; same file errors with the message above on `master`.

### DIFF
--- a/src/format/xml_db/entry.rs
+++ b/src/format/xml_db/entry.rs
@@ -57,6 +57,13 @@ pub struct Entry {
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub custom_data: Option<CustomData>,
+
+    /// KeePassXC extension: tracks the previous parent group when this entry
+    /// is moved between groups, used for undo/restore. Captured here so the
+    /// element is consumed by an explicit field rather than tripping the
+    /// strict serde deserializer.
+    #[serde(default, skip_serializing_if = "Option::is_none", with = "cs_opt_string")]
+    pub previous_parent_group: Option<UUID>,
 }
 
 impl Entry {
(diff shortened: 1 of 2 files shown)


sseemayer:
@jmylchreest Thanks for this contribution as well. #303 lands full support for KDBX4.1 databases, a superset of the changes of this commit.

PR #321, opened 2026-05-06
feat(entry): add attachments_named() iterator on EntryRef

### LINKED ISSUE #314: [v0.12] No public API to enumerate (name, attachment) pairs for an Entry
Hi Stefan! Following on from #262 closing with v0.12.

I'm building [diff-kdbx](https://github.com/queelius/diff-kdbx), a semantic diff for KDBX files (git textconv driver), and hit a wall while implementing attachment-aware diffs. Filing this in case it's a known wart you'd want a hand patching, or in case I'm just missing the right API.

## The problem

To diff two entries' attachments, I need to enumerate `(name, content)` pairs per entry. In v0.12 I can't do this with the public API:

- `Entry.attachments: HashMap<String, AttachmentId>` is `pub(crate)`.
- `Attachment.entries: HashMap<EntryId, String>` is `pub(crate)`.
- `EntryRef::attachments() -> impl Iterator<Item = AttachmentRef>` exists, but the items don't carry the per-entry name.
- `EntryRef::attachment_by_name(name)` exists, but only if you already know the names.

The only way to recover names from outside the crate is to round-trip through the `serialization` feature: serialize the entry to JSON, parse out the `attachments` object, then join with `Database::iter_all_attachments()` by ID. Which is what diff-kdbx currently does ([code here](https://github.com/queelius/diff-kdbx/blob/main/src/compute.rs)). It works but is obviously a workaround.

## Concrete use case

KDBX entries can have attachments like recovery codes, photos of ID, signed contracts. When diffing two snapshots of a vault, "attachment 'recovery.txt' was modified" is much more useful output than "an attachment with internal ID 7 was modified". The names are user-facing and stable.

## Suggested API addition

Something like one of:

```rust
impl<'a> EntryRef<'a> {
    /// Iterate over (filename, attachment) pairs for this entry.
    pub fn attachments_named(&self) -> impl Iterator<Item = (&str, AttachmentRef<'a>)> { ... }
}
```

or, more minimally:

```rust
impl<'a> EntryRef<'a> {
    /// Names of attachments referenced by this entry, in stable order.
    pub fn attachment_names(&self) -> impl Iterator<Item = &str> { ... }
}
```

Either gives downstream consumers a clean way to pair names with content. The first is more ergonomic; the second is the smallest possible surface change.

## Happy to PR

If you'd like, I'm happy to put up a PR. I have the diff-kdbx use case as a real-world consumer to validate the API against, so I can iterate on the shape until it feels right.

Two related (smaller) gaps I noticed while building, in case they're useful context (not asking you to do anything about them, just noting):

1. `History.entries: Vec<Entry>` is `pub(crate)`, but `History::get_entries() -> &Vec<Entry>` exists. So history is fine; just inconsistent with `Entry.attachments` which has no parallel public accessor.
2. Entry/Group `Uuid`s appear to be set internally on construction with no public setter. This makes deterministic test fixtures hard (each `Database::new()` produces fresh UUIDs). Workable, but a `with_uuid(uuid)` builder method would help downstream test infrastructure.

Thanks for v0.12 by the way: `iter_all_attachments()` and the `EntryMut::set_unprotected/set_protected` builders are nice. The new write API is much cleaner than I expected.

Alex (queelius)

queelius:
Closes #314.

Adds `EntryRef::attachments_named()`, returning an iterator of `(&str, AttachmentRef)` pairs. Mirrors the existing `attachments()` shape but yields the per-entry filename alongside each `AttachmentRef`, so callers don't need a second lookup via `attachment_by_name`.

## Why

`AttachmentRef` doesn't expose the per-entry filename (the key under which an attachment is stored on the entry). To get a `(name, data)` listing today, callers either:
- know the names ahead of time and call `attachment_by_name` per name, or
- enable the `serialization` feature and serialize the entry through JSON to extract the names.

Both are heavier than the natural iterator pattern.

## Implementation

11 lines on `EntryRef`. Uses `self.attachments.iter()` (HashMap iteration), maps each `(name, &id)` into `(name.as_str(), AttachmentRef::new(self.database, id))`. `AttachmentId` is `Copy`, so no clone needed.

## Test

Extends the existing `format::kdbx4::kdbx4_tests::test_attachments` roundtrip (which already creates two attachments with known data). Adds an inline iteration that asserts each `(name, AttachmentRef)` pair returns the expected bytes and that all names are unique.

Verified locally:
- `cargo build` clean
- `cargo test --all-features` all pass
- `cargo clippy --all-features` clean
- `cargo fmt --check` clean

## Notes

- Purely additive; no public API removed or changed.
- Iteration order is HashMap-based, matching `attachments()`.
- This is scoped to the `attachments_named()` part of #314. The UUID-setter / `History.entries` parts are separate (still discussing the consistency-check boundary on the issue thread).

### DIFF
--- a/src/db/types/entry.rs
+++ b/src/db/types/entry.rs
@@ -230,6 +230,17 @@ impl EntryRef<'_> {
             .map(move |attachment_id| AttachmentRef::new(self.database, attachment_id))
     }
 
+    /// Get an iterator over the (name, attachment) pairs of this entry.
+    ///
+    /// Useful when callers need both the attachment's filename (the key under
+    /// which it is stored on the entry) and its data, since [`AttachmentRef`]
+    /// itself does not expose the per-entry name.
+    pub fn attachments_named(&self) -> impl Iterator<Item = (&str, AttachmentRef<'_>)> {
+        self.attachments.iter().map(move |(name, &attachment_id)| {
+            (name.as_str(), AttachmentRef::new(self.database, attachment_id))
+        })
+    }
+
     /// Get the custom icon of this entry, if it exists and is a custom icon.
     pub fn custom_icon(&self) -> Option<CustomIconRef<'_>> {
         if let Some(Icon::Custom(custom_icon_id)) = self.icon {
--- a/src/format/kdbx4/mod.rs
+++ b/src/format/kdbx4/mod.rs
@@ -230,6 +230,20 @@ mod kdbx4_tests {
             entry.attachment_by_name("file2.txt").unwrap().get(),
             &[0x04, 0x03, 0x02, 0x01]
         );
+
+        // attachments_named() yields (name, AttachmentRef) pairs, allowing callers
+        // to recover the per-entry filename without a separate lookup.
+        let mut seen = std::collections::HashSet::new();
+        for (name, attachment) in entry.attachments_named() {
+            let expected: &[u8] = match name {
+                "file1.txt" => &[0x01, 0x02, 0x03, 0x04],
+                "file2.txt" => &[0x04, 0x03, 0x02, 0x01],
+                _ => panic!("unexpected attachment name: {}", name),
+            };
+            assert_eq!(attachment.get(), expected);
+            assert!(seen.insert(name.to_owned()), "duplicate name: {}", name);
+        }
+        assert_eq!(seen.len(), 2);
     }
 
     #[test]


sseemayer (approved):

sseemayer:
thanks for your contribution! 🚀

PR #323, opened 2026-05-08
test: add clean-room kdbx spec test suite

christhomas:
## Summary

Adds a kdbx4 spec test suite under `tests/`: round-trip across the cipher × KDF × compression × inner-stream × keyfile matrix, parser-error coverage for malformed inputs, keyfile format coverage (raw/hex/xml-v1/xml-v2), and binary-attachment round-tripping.

Fixtures are generated at test time from a seeded RNG — no on-disk corpus. All KDFs run with `iterations=1` per [CONTRIBUTING.md](../CONTRIBUTING.md). The suite is gated behind the `save_kdbx4` feature.

`tempfile`, `rand`, and `test-with` are added as native-only dev-dependencies.

## Test plan

- [x] `cargo test --features save_kdbx4 --test spec_round_trip --test broken_files --test keyfile_formats --test binary_pool --test cross_tool` — all pass locally.
- [x] `cargo test --no-default-features` — clean compile under cfg gate.

### Round 1 (33a5c50)

| Reviewer | Path:line | Verdict | Fix |
|---|---|---|---|
| sseemayer | tests/common/mod.rs:305 | adopt | `config_and_key_for` split into `Combo::get_config()` + `Combo::get_key()`; `minimal_database` / `rich_database` moved to methods on `Combo` |
| sseemayer | tests/broken_files.rs:25 | partial | Dropped `catch_unwind`. Deterministic mutations match specific variants (`VersionParse(InvalidKDBXIdentifier/InvalidKDBXVersion)`, `Key(IncorrectKey)`); ambiguous ones still use `assert!(res.is_err())` |
| sseemayer | tests/cross_tool.rs:25 | adopt | `test-with` (executable feature only) added as native dev-dep; test gated with `#[test_with::executable(keepassxc-cli)]` |
| sseemayer | tests/keyfile_formats.rs:67 | adopt | `root_child_count` helper removed; replaced with `parsed.root().entries().count()` / `groups().count()` |
| sseemayer | tests/spec_round_trip.rs:37 | adopt | Minimal-matrix and `second_save_is_self_consistent` now use `assert_eq!(parsed, db)`. Rich-matrix keeps per-field asserts so failures name the offending field+combo |
| sseemayer | top-level (rebase / feature-gate) | adopt | Rebased on master (PR #325); `save_to_vec` and all five test files gated behind `save_kdbx4` |
| sseemayer | top-level (`keepass::test::fixtures`) | discussion | Replied — proposed landing this PR as test-only first, then a follow-up that re-exports `Combo` + builders under a `test-fixtures` cargo feature |

### DIFF
--- a/tests/binary_pool.rs
+++ b/tests/binary_pool.rs
@@ -1,25 +1,15 @@
+#![cfg(feature = "save_kdbx4")]
 #![forbid(unsafe_code)]
 
 mod common;
 
-use common::{combo_by_label, config_and_key_for};
-use keepass::{
-    config::DatabaseConfig,
-    db::{Database, Value},
-    DatabaseKey,
-};
+use common::combo_by_label;
+use keepass::db::{Database, Value};
 use rand::{rngs::StdRng, RngCore, SeedableRng};
 
-fn baseline_setup() -> (DatabaseConfig, DatabaseKey, impl Fn() -> DatabaseKey) {
-    let combo = combo_by_label("aes256+none+inner-chacha20+argon2d");
-    let (cfg, key) = config_and_key_for(&combo);
-    let reopen_key = move || config_and_key_for(&combo).1;
-    (cfg, key, reopen_key)
-}
-
 fn drive(label: &str, payloads: Vec<Vec<u8>>) {
-    let (cfg, key, reopen_key) = baseline_setup();
-    let mut db = Database::with_config(cfg);
+    let combo = combo_by_label("aes256+none+inner-chacha20+argon2d");
+    let mut db = Database::with_config(combo.get_config());
 
     {
         let mut root = db.root_mut();
@@ -30,8 +20,8 @@ fn drive(label: &str, payloads: Vec<Vec<u8>>) {
         }
     }
 
-    let bytes = common::save_to_vec(&db, key);
-    let parsed = Database::open(&mut bytes.as_slice(), reopen_key())
+    let bytes = common::save_to_vec(&db, combo.get_key());
+    let parsed = Database::open(&mut bytes.as_slice(), combo.get_key())
         .unwrap_or_else(|err| panic!("{}: reopen failed: {:?}", label, err));
 
     assert_eq!(
--- a/tests/binary_pool.rs
+++ b/tests/binary_pool.rs
@@ -1,0 +1,98 @@
+#![forbid(unsafe_code)]
+
+mod common;
+
+use common::{combo_by_label, config_and_key_for};
+use keepass::{
+    config::DatabaseConfig,
+    db::{Database, Value},
+    DatabaseKey,
+};
+use rand::{rngs::StdRng, RngCore, SeedableRng};
+
+fn baseline_setup() -> (DatabaseConfig, DatabaseKey, impl Fn() -> DatabaseKey) {
+    let combo = combo_by_label("aes256+none+inner-chacha20+argon2d");
+    let (cfg, key) = config_and_key_for(&combo);
+    let reopen_key = move || config_and_key_for(&combo).1;
+    (cfg, key, reopen_key)
+}
+
+fn drive(label: &str, payloads: Vec<Vec<u8>>) {
+    let (cfg, key, reopen_key) = baseline_setup();
+    let mut db = Database::with_config(cfg);
+
+    {
+        let mut root = db.root_mut();
+        let mut e = root.add_entry();
+        e.set_unprotected("Title", "attachments");
+        for (i, content) in payloads.iter().enumerate() {
+            e.add_attachment(format!("blob-{i}.bin"), Value::Unprotected(content.clone()));
+        }
+    }
+
+    let bytes = common::save_to_vec(&db, key);
+    let parsed = Database::open(&mut bytes.as_slice(), reopen_key())
+        .unwrap_or_else(|err| panic!("{}: reopen failed: {:?}", label, err));
+
+    assert_eq!(
+        parsed.num_attachments(),
+        payloads.len(),
+        "{}: total attachments count",
+        label
+    );
+
+    let root = parsed.root();
+    let entry = root
+        .entries()
+        .next()
+        .unwrap_or_else(|| panic!("{}: no entry under root", label));
+
+    for (i, expected) in payloads.iter().enumerate() {
+        let name = format!("blob-{i}.bin");
+        let att = entry
+            .attachment_by_name(&name)
+            .unwrap_or_else(|| panic!("{}: missing attachment {}", label, name));
+        assert_eq!(
+            att.data.get().as_slice(),
+            expected.as_slice(),
+            "{}: payload {} differs",
+            label,
+            i
+        );
+    }
+}
+
+#[test]
+fn binary_pool_sizes_zero_one_sixteen() {
+    drive("sizes-tiny", vec![Vec::new(), vec![0x00], (0u8..16).collect()]);
+}
+
+#[test]
+fn binary_pool_4kib_random() {
+    let mut buf = vec![0u8; 4 * 1024];
+    StdRng::seed_from_u64(0x1234_5678).fill_bytes(&mut buf);
+    drive("sizes-4kib", vec![buf]);
+}
+
+#[test]
+fn binary_pool_1mib_random() {
+    let mut buf = vec![0u8; 1024 * 1024];
+    StdRng::seed_from_u64(0x9876_5432).fill_bytes(&mut buf);
+    drive("sizes-1mib", vec![buf]);
+}
+
+#[test]
+fn binary_pool_byte_patterns() {
+    drive(
+        "byte-patterns",
+        vec![
+            vec![0u8; 256],
+            vec![0xFFu8; 256],
+            (0..=255u8).collect(),
+            vec![
+                0xC3, 0x28, 0xA0, 0xA1, 0xE2, 0x28, 0xA1, 0xE2, 0x82, 0x28, 0xF0, 0x28, 0x8C, 0xBC, 0xF0, 0x90,
+                0x28, 0xBC, 0xF0, 0x28, 0x8C, 0x28,
+            ],
+        ],
+    );
+}
--- a/tests/broken_files.rs
+++ b/tests/broken_files.rs
@@ -1,34 +1,34 @@
+#![cfg(feature = "save_kdbx4")]
 #![forbid(unsafe_code)]
 
 mod common;
 
-use std::panic::{catch_unwind, AssertUnwindSafe};
-
-use common::{combo_by_label, config_and_key_for, fast_combo, minimal_database, DEMO_PASSWORD};
+use common::{combo_by_label, fast_combo, DEMO_PASSWORD};
+use keepass::error::{DatabaseKeyError, DatabaseOpenError, DatabaseVersionParseError};
 use keepass::{Database, DatabaseKey};
 
 fn baseline_blob() -> (Vec<u8>, DatabaseKey) {
     let combo = fast_combo();
-    let (cfg, key) = config_and_key_for(&combo);
-    let db = minimal_database(cfg);
-    let bytes = common::save_to_vec(&db, key);
-    (bytes, config_and_key_for(&combo).1)
-}
-
-fn assert_clean_error(blob: &[u8], key: DatabaseKey, label: &str) -> Result<(), String> {
-    let res = catch_unwind(AssertUnwindSafe(|| Database::open(&mut &blob[..], key)));
-    match res {
-        Ok(Ok(_)) => Err(format!("{label}: parser returned Ok on broken input")),
-        Ok(Err(_)) => Ok(()),
-        Err(_) => Err(format!("{label}: parser PANICKED on broken input")),
-    }
+    let db = combo.minimal_database();
+    let bytes = common::save_to_vec(&db, combo.get_key());
+    (bytes, combo.get_key())
 }
 
 #[test]
 fn mutation_bad_magic() {
     let (mut blob, key) = baseline_blob();
     blob[0] ^= 0xFF;
-    assert_clean_error(&blob, key, "BadMagic").unwrap();
+    let res = Database::open(&mut &blob[..], key);
+    assert!(
+        matches!(
+            res,
+            Err(DatabaseOpenError::VersionParse(
+                DatabaseVersionParseError::InvalidKDBXIdentifier
+            ))
+        ),
+        "BadMagic: unexpected result: {:?}",
+        res
+    );
 }
 
 #[test]
@@ -36,7 +36,17 @@ fn mutation_bad_kdbx_version() {
     let (mut blob, key) = baseline_blob();
     blob[10] = 5;
     blob[11] = 0;
-    assert_clean_error(&blob, key, "BadKdbxVersion").unwrap();
+    let res = Database::open(&mut &blob[..], key);
+    assert!(
+        matches!(
+            res,
+            Err(DatabaseOpenError::VersionParse(
+                DatabaseVersionParseError::InvalidKDBXVersion { .. }
+            ))
+        ),
+        "BadKdbxVersion: unexpected result: {:?}",
+        res
+    );
 }
 
 #[test]
@@ -44,14 +54,16 @@ fn mutation_truncated_at_header() {
     let (blob, key) = baseline_blob();
     let cut = (blob.len() / 4).clamp(20, 80);
     let truncated = &blob[..cut];
-    assert_clean_error(truncated, key, "TruncatedAtHeader").unwrap();
+    let res = Database::open(&mut &truncated[..], key);
+    assert!(res.is_err(), "TruncatedAtHeader: expected Err, got Ok");
 }
 
 #[test]
 fn mutation_truncated_at_payload() {
     let (blob, key) = baseline_blob();
     let truncated = &blob[..blob.len() - 8];
-    assert_clean_error(truncated, key, "TruncatedAtPayload").unwrap();
+    let res = Database::open(&mut &truncated[..], key);
+    assert!(res.is_err(), "TruncatedAtPayload: expected Err, got Ok");
 }
 
 #[test]
@@ -59,7 +71,8 @@ fn mutation_bad_header_sha256() {
     let (mut blob, key) = baseline_blob();
     let header_end = locate_header_end(&blob).expect("locate header end");
     blob[header_end] ^= 0x01;
-    assert_clean_error(&blob, key, "BadHeaderSha256").unwrap();
+    let res = Database::open(&mut &blob[..], key);
+    assert!(res.is_err(), "BadHeaderSha256: expected Err, got Ok");
 }
 
 #[test]
@@ -67,7 +80,8 @@ fn mutation_bad_header_hmac() {
     let (mut blob, key) = baseline_blob();
     let header_end = locate_header_end(&blob).expect("locate header end");
     blob[header_end + 32 + 1] ^= 0x80;
-    assert_clean_error(&blob, key, "BadHeaderHmac").unwrap();
+    let res = Database::open(&mut &blob[..], key);
+    assert!(res.is_err(), "BadHeaderHmac: expected Err, got Ok");
 }
 
 #[test]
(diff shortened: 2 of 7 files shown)


christhomas:
I noticed that the code was not passing the cargo fmt part of the pipeline. So I proactively updated the commit with cargo fmt first

christhomas:
I also scrubbed and cleaned the commits to remove unnecessary information, the kind of thing I would do for PRs done at my company to make it more digestable

tests/common/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+//!
+//! All fixtures are produced programmatically from a seeded `StdRng` and the
+//! library's own writers. No data is copied from KeePassXC, KeePass2,
+//! KeePassDX, pykeepass, or any other GPL'd project; the entire corpus is
+//! reproducible from this source file.
+//!
+//! Ported to the keepass = 0.12.5 API: attachments are now first-class
+//! database-owned objects (no more `header_attachments`), entries are added
+//! through `GroupMut::add_entry()` returning `EntryMut`, and the old
+//! `db.root.add_child(Node)` / `Entry::new()` constructors are gone.
```
sseemayer: Can you please remove this and other comments that are related to the history of the code base? they are relevant to your main project, but not the keepass crate.

tests/common/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    // 3) Password + raw keyfile composition.
+    combos.push(Combo {
+        label: "aes256+gz+inner-chacha20+argon2d+pw+kf-raw32",
+        outer_cipher: OuterCipherConfig::AES256,
+        compression: CompressionConfig::GZip,
+        inner_cipher: InnerCipherConfig::ChaCha20,
+        kdf: fast_argon2(),
+        master_key: MasterKey::PasswordAndKeyfile(
+            "correct horse battery staple",
```
sseemayer: please use `demopass` as a password wherever possible to simplify future debugging work.

tests/common/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+    (cfg, key)
+}
+
+/// Construct a "rich" test database: 1 root + 10 entries, custom fields,
+/// binary attachments (small + medium + non-UTF-8), tags, expiry, custom data,
+/// and a recycle-bin group.
+///
+/// In keepass 0.12.5 the root group's children are iterated as separate
+/// entry/group iterators. The 10 entries are direct children of root; the
+/// recycle bin is a child group.
```
sseemayer: another instance of a version history comment to be removed.

tests/broken_files.rs
```diff
@@ -0,0 +1,332 @@
...
+//! Every fixture is generated programmatically from a healthy seeded vault,
+//! then mutated by category. No file fetched at test time, no GPL'd inputs.
+//!
+//! Ported to keepass 0.12.5: `Entry::new()` and `db.root.add_child(...)` are
+//! gone, and entries no longer carry a public `uuid` field. We construct
+//! duplicate-UUID fixtures by reaching into `Database::entries` directly,
+//! which we can do because the test crate is not the keepass crate (so we
+//! can't actually reach the private map; instead we serialise → mutate XML).
+//! Where the new model genuinely cannot express the broken shape pre-encrypt,
+//! the tests stay `#[ignore]`'d with an updated rationale.
```
sseemayer: the comments about fixture generation, and no presence of GPL'd inputs is unnecessary, just like the version history.

tests/broken_files.rs
```diff
@@ -0,0 +1,332 @@
...
+//     would be needed.
+//
+//   * Dangling `<Binary Ref="N"/>` to a missing pool entry: in 0.12.5 the
+//     pool is internal and `add_attachment` always creates a valid reference,
+//     so we can't construct this case from the public API at all. The
+//     fundamental scenario also no longer applies — there is no separate
+//     "binary pool" the entry references; attachments are the same objects.
+//
+// All of these mutations therefore stay `#[ignore]`'d with updated rationales.
+// ---------------------------------------------------------------------------
```
sseemayer: please remove the explanation of removed tests

tests/broken_files.rs
```diff
@@ -0,0 +1,332 @@
...
+#[ignore = "tracked: missing UUID requires direct XML injection — the model auto-populates GroupId on add_group()"]
+fn mutation_missing_group_uuid() {}
+
+#[test]
+#[ignore = "no longer applicable in keepass 0.12.5: attachments are first-class database objects, not a separate 'binary pool' the entry references; dangling Ref scenarios cannot be constructed through the public API and the underlying behaviour they tested no longer exists"]
+fn mutation_dangling_binary_ref() {}
+
+#[test]
+#[ignore = "tracked: requires injection of broken XML inside the encrypted payload"]
+fn mutation_bad_xml_inner_header() {}
```
sseemayer: there is no point in having empty ignored tests around.

tests/cross_tool.rs
```diff
@@ -0,0 +1,75 @@
...
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("spawn keepassxc-cli");
+    child
+        .stdin
+        .as_mut()
+        .unwrap()
+        .write_all(b"correct horse battery staple\n")
```
sseemayer: please use `demopass` as a password

tests/keyfile_formats.rs
```diff
@@ -0,0 +1,108 @@
...
+    let r = db.root();
+    r.entries().count() + r.groups().count()
+}
+
+fn drive(kind: KeyfileKind, label: &str) {
+    let combos = round_trip_combos();
+    // pick the keyfile-only combo for a deterministic key
+    let combo = combos
+        .iter()
+        .find(|c| c.label.contains("aes256+gz+inner-chacha20+argon2d+kf-raw32"))
```
sseemayer: it would be cleaner to have a separate utility function to return only the baseline combo.

tests/spec_round_trip.rs
```diff
@@ -0,0 +1,228 @@
...
+//! Round-trip the full kdbx4 spec matrix against the library's own writer.
+//!
+//! Each combo (cipher × KDF × compression × inner-stream × master-key) is
+//! generated programmatically, saved to `Vec<u8>`, then reopened and
+//! compared semantically. No on-disk fixtures, no GPL'd test data.
+//!
+//! Ported to keepass 0.12.5: there is no longer a `db.root.children: Vec<Node>`
+//! tree, and there is no separate `header_attachments` collection — we
+//! traverse via `root().entries()` / `root().groups()` and inspect attachments
+//! through `iter_all_attachments()` / `entry.attachment_by_name(..)`.
```
sseemayer: another unnecessary version note

tests/spec_round_trip.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+#[test]
+fn second_save_is_self_consistent() {
+    // For one well-tested combo, save → reopen → save → reopen → compare.
+    let combos = round_trip_combos();
+    let combo = combos
+        .iter()
+        .find(|c| c.label == "aes256+gz+inner-chacha20+argon2d")
+        .expect("baseline combo present");
```
sseemayer: could be replaced with a utility function returning a baseline combo.

tests/spec_round_trip.rs
```diff
@@ -0,0 +1,228 @@
...
+    );
+}
+
+#[test]
+fn protected_field_decrypts_after_round_trip() {
+    let combos = round_trip_combos();
+    let combo = combos
+        .iter()
+        .find(|c| c.label == "aes256+gz+inner-chacha20+argon2d")
+        .expect("baseline combo present");
```
sseemayer: could be replaced with a utility function returning a baseline combo.

tests/binary_pool.rs
```diff
@@ -0,0 +1,116 @@
...
+};
+use rand::{rngs::StdRng, RngCore, SeedableRng};
+
+fn baseline_setup() -> (DatabaseConfig, DatabaseKey, impl Fn() -> DatabaseKey) {
+    let combos = round_trip_combos();
+    let combo = combos
+        .iter()
+        .find(|c| c.label == "aes256+none+inner-chacha20+argon2d")
+        .expect("baseline combo")
+        .clone();
```
sseemayer: could be replaced with a utility function returning a baseline combo.

tests/broken_files.rs
```diff
@@ -0,0 +1,332 @@
...
+    MissingEntryUuid,
+    MissingGroupUuid,
+    NoRootGroup,
+    MultipleRootGroups,
+    EmptyXml,
+    MalformedUtf8,
+    WrongPassword,
+    KeyfileMismatch,
+    DataAfterEnd,
+}
```
sseemayer: This enum is not used anywhere

sseemayer (commented):
thank you for your PR. I generally like the idea of additional matrix testing of various settings, but I have a lot of questions about the code and approach.

1. I have never heard the term "clean-room spec test suite" before. What is a spec test, and what makes it clean-room? What specification are you testing against? Can you provide a source?
2. I am not sure what specific error cases these tests are supposed to protect against, in addition to the tests that are already there. Why a new test suite instead of amending the existing tests?
3. The combos being explored are mostly related to the outer and inner cipher suites and key derivation function. Why matrix-test whether changes to the outer cipher suites have an effect on the content of the inner XML database? 
4. I do not get your point about integration tests not getting installed into crates using this crate as a dependency, why would other crates want to have "integration coverage"?

Overall, the tests are very hard to read, and there are several left-over comments, pointless tests and unused dead code. I have added comments to some of the regions that stand out to me, but this is a non-exhaustive list.

droidmonkey:
Omg I love this, I wanted to convert keepassxc tests to be deterministic kdbx on run time instead of prepared databases.

christhomas:
> Omg I love this, I wanted to convert keepassxc tests to be deterministic kdbx on run time instead of prepared databases.

I'm birtish, and this sounds like sarcasm, so I'm confused :D 

sseemayer actually made a bunch of good points, some stuff was too enthusiastic, so I pulled it back a bit and removed references to my project where it shouldn't have been there, etc.

droidmonkey:
Not sarcastic at all, truly meant it

christhomas:
@droidmonkey if you have any requests, things you want to expand on to make it better, then go ahead and ask. I'm open for suggestions

tests/common/mod.rs
christhomas: addresses in the new commit (I'll close the other conversations with no comment because it's the same as this one)

tests/common/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    db
+}
+
+pub fn minimal_database(config: DatabaseConfig) -> Database {
+    let mut db = Database::with_config(config);
+    let mut root = db.root_mut();
+    let mut e = root.add_entry();
+    e.set_unprotected("Title", "only-entry");
+    db
+}
```
sseemayer: The `config_and_key_for`, `minimal_database` and `rich_database` functions could be put into `impl Combo`, potentially splitting `config_and_key_for` into `get_config` and `get_key` methods.  That would lead to a very nice API, e.g.:

```rust
#[test]
fn matrix_round_trip_minimal_database() {
    for combo in &round_trip_combos() {
        let db = combo.minimal_database();
        let bytes = common::save_to_vec(&db, combo.get_key());
        assert!(
            bytes.len() > 32,
            "combo {} produced suspiciously small output ({} bytes)",
            combo.label,
            bytes.len()
        );
        assert_eq!(
            &bytes[..4],
            &[0x03, 0xd9, 0xa2, 0x9a],
            "combo {} missing kdbx magic",
            combo.label
        );
        let parsed = Database::open(&mut bytes.as_slice(), combo.get_key())
            .unwrap_or_else(|e| panic!("combo {} reopen failed: {:?}", combo.label, e));
        assert_eq!(root_child_count(&parsed), root_child_count(&db));
    }
}
```

tests/broken_files.rs
```diff
@@ -0,0 +1,166 @@
...
+
+fn assert_clean_error(blob: &[u8], key: DatabaseKey, label: &str) -> Result<(), String> {
+    let res = catch_unwind(AssertUnwindSafe(|| Database::open(&mut &blob[..], key)));
+    match res {
+        Ok(Ok(_)) => Err(format!("{label}: parser returned Ok on broken input")),
+        Ok(Err(_)) => Ok(()),
+        Err(_) => Err(format!("{label}: parser PANICKED on broken input")),
+    }
+}
```
sseemayer: Tests will already fail if calling `Database::open` panics, so wrapping in `catch_unwind(AssertUnwindSafe(|| { ... }))` is unnecessary

Instead of checking whether any error whatsoever is returned, would it be possible to replace this with something like this:

```rust

#[test]
fn mutation_bad_magic() {
    let (mut blob, key) = baseline_blob();
    blob[0] ^= 0xFF;
    let res = Database::open(&mut &blob[..], key);
    assert_matches!(res, Err(DatabaseOpenError::VersionParse(DatabaseVersionParseError::InvalidKDBXIdentifier)));
}
```

tests/cross_tool.rs
```diff
@@ -0,0 +1,64 @@
...
+        .map(|s| s.success())
+        .unwrap_or(false)
+}
+
+#[test]
+fn keepassxc_cli_lists_our_vault() {
+    if !keepassxc_cli_present() {
+        eprintln!("keepassxc-cli not on PATH — skipping");
+        return;
+    }
```
sseemayer: Instead of silently skipping within the test when `keepassxc-cli` is not available, I would prefer to add a dev-dependency on [test_with](https://docs.rs/test-with/latest/test_with/) so that the test gets marked as ignored when the command is not available, e.g.:

```rust
#[test]
#[test_with::executable(keepassxc-cli)]
fn keepassxc_cli_lists_our_vault() {
    /* ... */
}
```

tests/keyfile_formats.rs
```diff
@@ -0,0 +1,83 @@
...
+    let key = DatabaseKey::new()
+        .with_keyfile(&mut Cursor::new(garbage.clone()))
+        .expect("garbage keyfile accepted");
+    let db = minimal_database(cfg);
+    let saved = common::save_to_vec(&db, key);
+    let reopen = DatabaseKey::new()
+        .with_keyfile(&mut Cursor::new(garbage))
+        .expect("garbage keyfile accepted 2");
+    let parsed = Database::open(&mut saved.as_slice(), reopen).expect("garbage keyfile reopens");
+    assert_eq!(root_child_count(&parsed), 1);
```
sseemayer: the `root_child_count` helper can be replaced with something like this:

```rust
assert_eq!(parsed.root().entries().count(), 1);
assert_eq!(parsed.root().groups().count(), 0);
```

tests/spec_round_trip.rs
```diff
@@ -0,0 +1,152 @@
...
+        );
+        assert_eq!(
+            &bytes[..4],
+            &[0x03, 0xd9, 0xa2, 0x9a],
+            "combo {} missing kdbx magic",
+            combo.label
+        );
+        let parsed = Database::open(&mut bytes.as_slice(), opening_key_for(combo))
+            .unwrap_or_else(|e| panic!("combo {} reopen failed: {:?}", combo.label, e));
+        assert_eq!(root_child_count(&parsed), root_child_count(&db));
```
sseemayer: `Database` implements `Eq`, so you can also use a much stronger check:

```rust
assert_eq!(parsed, db);
```

sseemayer (commented):
Thanks a lot for your cleanup! I have added some smaller comments but this is already looking very good. 

One general discussion topic: Do we want to expose the test fixture generation as part of the main crates' API, something like a `keepass::test::fixtures` module? You mentioned the use case of wanting to use these test fixtures from within other crates.

sseemayer:
I wanted to add a few commits to fix up the tests, but don't seem to have permissions to add to your branch, so here are hints instead:

* the test for cross-tool compatibility can be fixed by rebasing on the current master branch containing the #325 PR.
* you need to feature-gate the save_to_vec utility function and all your test files, e.g.
   ```rust
   pub fn save_to_vec(db: &Database, key: DatabaseKey) -> Vec<u8> {
   ``` 
   
   and at the top of each test file:
   
   ```rust
   #![cfg(feature = "save_kdbx4")]
   ```

tests/common/mod.rs
christhomas: Done in 33a5c50. `config_and_key_for` split into `Combo::get_config()` + `Combo::get_key()`, and `minimal_database` / `rich_database` are methods on `Combo`. Call sites match your snippet — `combo.minimal_database()` / `common::save_to_vec(&db, combo.get_key())`.

tests/broken_files.rs
christhomas: Done in 33a5c50. Dropped the `catch_unwind` wrapper — agreed, panics fail the test naturally. Tightened the deterministic mutations: `bad_magic` and `bad_kdbx_version` now match `DatabaseOpenError::VersionParse(..)`, and `wrong_password` / `keyfile_mismatch` match `DatabaseOpenError::Key(DatabaseKeyError::IncorrectKey)`. The truncation / bad-header / bad-cipher / bad-KDF mutations can route through several `DatabaseOpenError` variants depending on which validation fires first, so those still use `assert!(res.is_err())`. Happy to lock them down to specific variants if you want a particular ordering enforced.

tests/cross_tool.rs
christhomas: Done in 33a5c50. Added `test-with` (executable feature only) as a native dev-dep and gated `keepassxc_cli_lists_our_vault` with `#[test_with::executable(keepassxc-cli)]`. The test now reports as `ignored` rather than silently passing when the binary isn't on PATH.

tests/keyfile_formats.rs
christhomas: Done in 33a5c50. Replaced with `parsed.root().entries().count() == 1` and `parsed.root().groups().count() == 0`; the `root_child_count` helper is gone.

tests/spec_round_trip.rs
christhomas: Done in 33a5c50. The minimal-database matrix now uses `assert_eq!(parsed, db)`. Kept per-field assertions in the rich-database test so a regression names the failing field+combo rather than dumping the full `Database` struct, but `second_save_is_self_consistent` also benefits from the stronger `assert_eq!(parsed_a, parsed_b)`.

christhomas:
Pushed 33a5c50 addressing the inline review feedback. Highlights:

- Rebased on master so `cross_tool` picks up #325.
- `save_to_vec` + all five test files are gated behind `save_kdbx4` (`#![cfg(feature = "save_kdbx4")]` at file top, `#[cfg]` on the helper).
- `cross_tool` now uses `#[test_with::executable(keepassxc-cli)]` — reports as `ignored` instead of silently passing when the binary isn't on PATH.

For the `keepass::test::fixtures` idea — happy to do it; my preference would be to land this as test-only first, then a follow-up that re-exports `Combo` + the builders under a `test-fixtures` cargo feature so downstream crates can opt in without paying the cost in default builds. Let me know if you'd rather bundle it into this PR.

sseemayer:
thanks a lot for your quick responses!

> my preference would be to land this as test-only first, then a follow-up that re-exports `Combo` + the builders under a `test-fixtures` cargo feature so downstream crates can opt in without paying the cost in default builds. Let me know if you'd rather bundle it into this PR.

sounds good!

christhomas:
@sseemayer and what would you like me to do considering @droidmonkey's comment about the direction of the test suite, should I try to expand on this idea? do you have any opinion on what he said?

sseemayer:
> @sseemayer and what would you like me to do considering @droidmonkey's comment about the direction of the test suite, should I try to expand on this idea? do you have any opinion on what he said?

right now, the crate's test suite is very chaotic because it grew historically. I would like to look at the totality of currently existing tests and clearly categorize them as such:

  1. unit tests that live close to the various parts of the code -> keep them as-is
  2. integration tests for the KDBX4.1 format reading and writing -> move as many of those to your programmatically generated test fixtures as possible
  3. file format support tests that take the currently existing fixture files and ensure we continue to support reading databases of various other KDBX generators -> keep mostly as-is, there might be some chance for thinning out
  4. regression tests for specific error cases encountered in the past -> keep them as-is

sseemayer (approved):

christhomas:
@sseemayer I could look at this one if you like?

> integration tests for the KDBX4.1 format reading and writing -> move as many of those to your programmatically generated test fixtures as possible

it doesn't hurt to try and you can always reject/advise changes otherwise. it's just the time needed to review :P

sseemayer:
thanks for the offer, that would be appreciated! I have also created #328 to track this.

PR #325, opened 2026-05-10
fix: keepass and keepassxc compatibility

### LINKED ISSUE #324: KeePass and KeePassXC Client Compatibility?
I'm using KeePassXC 2.7.12 and KeePass 2.61 on Arch, and it seems like the default vaults created with this crate are not compatible with either client. When I run the save_database example and try to open the generated .kbdx file, KeePass says it can't read the vault due to malformed Base64 and KeePassXC says it can't read it due to an invalid number.

I've done a bit of digging into this, and it seems to be mostly an issue with the default values you use for the XML payload. I made some changes to a local copy of the repo and managed to get it generating vaults that could be opened by KeePass, and I feel pretty confident I could do the extra work to get it working with KeePassXC as well. I'm happy to do the work if you would want that PR, but I just want to check first so I don't waste my time working on something that wouldn't be wanted.

sseemayer:
fixes #324

the new XML serialization code emitted empty optional fields that would lead to parse errors with keepass and keepassxc. Also, keepass seems to accept only base64-encoded timestamps.

### DIFF
--- a/src/format/xml_db/entry.rs
+++ b/src/format/xml_db/entry.rs
@@ -22,22 +22,32 @@ pub struct Entry {
     #[serde(rename = "UUID")]
     pub uuid: UUID,
 
-    #[serde(default, rename = "IconID", with = "cs_opt_fromstr")]
+    #[serde(
+        default,
+        rename = "IconID",
+        with = "cs_opt_fromstr",
+        skip_serializing_if = "Option::is_none"
+    )]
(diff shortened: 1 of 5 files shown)


varjolintu (dismissed):

varjolintu (approved):

sseemayer:
thanks again to @SatelliteDish for help fixing this quickly! 🚀

PR #329, opened 2026-05-11
feat(db): pub add_entry_with_id / add_group_with_id with Result return (closes #322)

### LINKED ISSUE #322: Add possibility to add entries / groups with specified identifiers
Original discussion in #314 - splitting into a separate issue.

Developers that use the keepass crate might want to write tests where entries, groups, attachments, custom icons are constructed with defined UUIDs. This use case is currently not supported as the database hands out identifiers through `Group::add_entry`, et al. , to ensure that the invariants of the database object structure (references always point to objects that exist, disallow several objects with the same identifier, no circular groups, etc.) are maintained.

We should add a safe API to add entries, groups, attachments, and custom icons with specific identifiers.

queelius:
Closes #322.

Implements the shape you proposed in the issue thread: `Group::add_entry_with_id` and `Group::add_group_with_id` become `pub` and return `Result<_, AddError>` instead of panicking on duplicate UUIDs. `EntryId::from_uuid` and `GroupId::from_uuid` are now `pub`, and `From<Uuid>` is implemented for both.

## What

- `Group::add_entry_with_id(EntryId) -> Result<EntryMut<'_>, AddError>` (was `pub(crate)`, panicked on dup)
- `Group::add_group_with_id(GroupId) -> Result<GroupMut<'_>, AddError>` (same)
- `EntryId::from_uuid` and `GroupId::from_uuid` are now `pub`
- `impl From<Uuid> for EntryId` and `impl From<Uuid> for GroupId`
- New `AddError` enum (re-exported from `db::types`) with `DuplicateEntryUuid(Uuid)` and `DuplicateGroupUuid(Uuid)` variants. Marked `#[non_exhaustive]` so future variants (custom icons, attachments) can be added without breaking semver.

The doc examples in the new methods spell out the call shape you sketched in #322:

```rust
let entry_id: EntryId = uuid!("00000000-0000-0000-0000-000000000001").into();
db.root_mut()
    .add_entry_with_id(entry_id)
    .unwrap()
    .edit(|e| {
        e.set_unprotected(fields::TITLE, "My entry with defined UUID");
    });
```

## Internal callers

- `Group::add_entry` and `Group::add_group` keep their existing infallible signatures. They generate a fresh v4 UUID and unwrap the new `Result` with `.expect` on the "fresh `Uuid::new_v4()` cannot collide" invariant, matching the existing `allow(clippy::expect_used)` style used elsewhere for unreachable cases.
- `merge.rs` unwraps with `.expect` at the two call sites where the merge algorithm has already filtered the id through `source.difference(dest)`.
- The XML loader threads `AddError` through `UnprotectError` via a new `DuplicateIdentifier` variant, so a malformed KDBX with duplicate UUIDs now surfaces a structured error instead of a panic.

## Scope

CustomIcon and attachment cases are NOT in this PR. You flagged attachments as needing special logic (integer ids from the XML pool); CustomIcon does not currently expose an analogous `add_with_id` constructor so there was no panic to remove. Happy to follow up on either with separate PRs once you confirm the shape.

## Tests

Five new tests in `src/db/types/group.rs`:

- `add_entry_with_id_uses_supplied_uuid`
- `add_entry_with_id_duplicate_returns_error`
- `add_group_with_id_uses_supplied_uuid`
- `add_group_with_id_duplicate_returns_error`
- `from_uuid_impls_match_constructors`

Plus the doc tests for the new methods. `cargo test --all-features` passes locally; `cargo fmt --check` clean; `cargo doc` clean.

## API compatibility

The only public-surface signatures that changed are the two `_with_id` methods (newly `pub`) and `UnprotectError` gaining a `#[non_exhaustive]` variant. `Group::add_entry()` and `Group::add_group()` are unchanged.


src/db/types/group.rs
```diff
@@ -543,6 +597,22 @@ impl GroupMut<'_> {
...
+pub enum AddError {
+    /// An entry with the given UUID already exists in the database.
+    #[error("entry with UUID {0} already exists")]
+    DuplicateEntryUuid(Uuid),
+
+    /// A group with the given UUID already exists in the database.
+    #[error("group with UUID {0} already exists")]
+    DuplicateGroupUuid(Uuid),
+}
+
```
sseemayer: Can you anticipate other error types that could occur for `add_*_with_id`? If no, instead of an enum error type, two struct error types `pub struct DuplicateEntryIdError(EntryId)` and `DuplicateGroupIdError(GroupId` would be cleaner because an enum would force the user to match for variants that can not occur.

sseemayer (commented):
@queelius Thanks a lot for the PR - it already looks great! 🚀

One comment regarding error types.

Please also make sure to re-export the new error types in `src/error.rs`

sseemayer (changes requested):

sseemayer (dismissed):

sseemayer:
@louib @varjolintu @droidmonkey a review would be appreciated.

src/db/merge.rs
```diff
@@ -157,7 +157,10 @@ fn merge_groups(dest_db: &mut Database, source_db: &Database, log: &mut MergeLog
         // does the parent exist in dest?
         if let Some(mut parent) = dest_db.group_mut(parent_id) {
             // yes - re-add the group
-            let mut dest_group = parent.add_group_with_id(id);
+            #[allow(clippy::expect_used)] // id was selected from source_groups.difference(dest_groups)
+            let mut dest_group = parent
+                .add_group_with_id(id)
+                .expect("group id was filtered to be absent in dest");
```
droidmonkey: Filtered? I believe this would be something like "Group ID found in dest deleted objects registry". If not, I dont understand this comment.

src/db/merge.rs
```diff
@@ -395,7 +398,10 @@ fn merge_entries(dest_db: &mut Database, source_db: &Database, log: &mut MergeLo
             continue;
         };
 
-        let mut entry = parent.add_entry_with_id(id);
+        #[allow(clippy::expect_used)] // id was selected from source_entries.difference(dest_entries)
+        let mut entry = parent
+            .add_entry_with_id(id)
+            .expect("entry id was filtered to be absent in dest");
```
droidmonkey: Same comment here

src/db/types/group.rs
```diff
@@ -315,45 +326,88 @@ impl GroupMut<'_> {
...
-        self.groups.insert(id);
-        self.database.groups.insert(id, new_group);
-
-        GroupMut::new(self.database, id)
+        // A freshly-generated v4 UUID does not collide with an existing identifier in any
+        // realistic scenario, so the duplicate check inside `add_group_with_id` cannot trip
+        // here.
+        #[allow(clippy::expect_used)] // fresh v4 UUID cannot collide
+        self.add_group_with_id(id)
+            .expect("fresh v4 UUID cannot collide with an existing group identifier")
```
droidmonkey, line 335: You say two different error strings with the same return exception. Merge assumes deleted object and here you assume same uuid.

src/db/types/group.rs
```diff
@@ -315,45 +326,88 @@ impl GroupMut<'_> {
...
-        self.entries.insert(id);
-        self.database.entries.insert(id, new_entry);
-
-        EntryMut::new(self.database, id)
+        // A freshly-generated v4 UUID does not collide with an existing identifier in any
+        // realistic scenario, so the duplicate check inside `add_entry_with_id` cannot trip
+        // here.
+        #[allow(clippy::expect_used)] // fresh v4 UUID cannot collide
+        self.add_entry_with_id(id)
+            .expect("fresh v4 UUID cannot collide with an existing entry identifier")
```
droidmonkey, line 349: Same here

droidmonkey (changes requested):

src/db/merge.rs
sseemayer: This code path is about a whole group hierarchy that was removed from `dest`, but needs to be re-added because a child entry or group was updated.

At this stage, the group to be re-added should not exist, because it was previously scanned and found missing.

I am adding a commit to reword the error message.

src/db/types/group.rs
sseemayer: this code path generates a fresh v4 UUID at random. It is very unlikely that a freshly generated UUID could collide with any previously-existing UUID.

src/db/types/group.rs
sseemayer: addressed above.

src/db/merge.rs
sseemayer: @droidmonkey  please have another look.

src/db/merge.rs
sseemayer: addressed with the same commit.

droidmonkey (approved):

PR #331, opened 2026-05-13
test: ensure that KeePassXC and KeePass can read our KDBX files

### LINKED ISSUE #327: Add keepassxc and kpscript to CI environment
#323 added a test with `keepassxc-cli` to ensure KeePassXC can parse the KDBX files generated by this crate, but this rule is currently ignored because `keepassxc-cli` is not available in the CI environment.

Install both `keepassxc` and `kpscript` [1] (to check whether KeePass can parse our KDBX files) so that the test is no longer ignored.

KPScript can be used to programmatically list a database with a command as such:

```bash
kpscript -c:ListEntries database.kdbx -pw:demopass
```

Sources:

[1] https://keepass.info/help/v2_dev/scr_index.html

sseemayer:
closes #327

Installs both KeePassXC and the original KeePass with the KPScript plugin as part of our CI/CD pipeline, so that our cross-tool tests can ensure that both KeePassXC and the original KeePass are able to load KDBX files generated by this crate.


tests/cross_tool.rs
```diff
@@ -8,6 +8,39 @@ use std::process::{Command, Stdio};
 
 use common::{baseline_combo, DEMO_PASSWORD};
 
+#[test_with::executable(kpscript)]
```
louib, line 13: @sseemayer if for whatever reason the executable is not found, we will simply not run this test, from what I understand from [the macro's documentation](https://docs.rs/test-with/latest/test_with/attr.executable.html). Is there a way that we can override this behavior, at least when running the tests in CI?

.github/workflows/merge.yml
```diff
@@ -15,6 +15,17 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
+      - name: Install cross-tool test dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y keepassxc mono-complete unzip
+          curl -fsSL "https://sourceforge.net/projects/keepass/files/KeePass%202.x/2.61.1/KeePass-2.61.1.zip/download" -o /tmp/KeePass.zip
+          curl -fsSL "https://keepass.info/extensions/v2/kpscript/KPScript-2.61.1.zip" -o /tmp/KPScript.zip
```
louib, line 23: @sseemayer this is not a huge deal since this is not happening in the job that actually publishes the package, but ideally we should be verifying the checksum of the zip file we download here.

tests/cross_tool.rs
sseemayer: it's not possible within `test-with`, but with [c5e8e3c](https://github.com/sseemayer/keepass-rs/pull/331/commits/c5e8e3c8e0e44d0f5bf5f04fce73f53617cf7569) I have added another CI/CD step that will fail unless the two programs are found.

.github/workflows/merge.yml
sseemayer: good idea, added in [5469656](https://github.com/sseemayer/keepass-rs/pull/331/commits/5469656675f4fada5c354295b23108539273a2d4)

sseemayer:
@louib please have another look, all your comments should now be addressed.

droidmonkey (approved):

droidmonkey:
Great idea!

PR #340, opened 2026-05-24
fix: invalid crate keywords

sseemayer:
the crate gets rejected by crates.io because two-word keywords such as "password manager" are disallowed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "keepass"
 description = "KeePass .kdbx database file parser"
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):
Can we use `password-manager`?

sseemayer:
> Can we use `password-manager`?

I did not want to risk it. Edit: up to 20 alphanumeric characters plus `_-+`, so it would be allowed, but I guess password is more inclusive.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field

PR #342, opened 2026-05-25
feat(merge): add Database::merge_with_ancestor for three-way merging

### LINKED ISSUE #337: feat: three-way merge variant (`Database::merge_with_ancestor`)
Same as #336, feel free to close if #335 is shaky and/or will be rewritten

`Database::merge(&mut self, other: &Database) -> Result<MergeLog, MergeError>` [here](https://github.com/sseemayer/keepass-rs/blob/89d033d1af5da5a89037a8cab0b1e6fa14094704/src/db/merge.rs#L60) is a two-way merge: conflict resolution uses `last_modification` timestamp comparison only.

This works for many sync scenarios, but cannot distinguish "both sides changed since the common ancestor" from "only one side changed." The former needs user attention; the latter can be auto-resolved without surfacing a conflict.

## Proposed API

```rust
pub fn merge_with_ancestor(
    &mut self,
    other: &Database,
    base: &Database,
) -> Result<MergeLog, MergeError>;
```

For each entry / group:

- Changed in `self` only: keep `self`'s value.
- Changed in `other` only: take `other`'s value.
- Changed identically in both: trivially equal.
- Changed differently in both: real conflict; surface via `MergeLog.warnings`, resolve by existing newest-wins policy (or a new explicit policy parameter).

The existing two-way `merge` can remain as a convenience shim:

```rust
pub fn merge(&mut self, other: &Database) -> Result<MergeLog, MergeError> {
    self.merge_with_ancestor(other, &Database::new())
}
```

## Use case

Cloud-synced .kdbx databases (Dropbox, iCloud, etc.) where the last-known-synced version on disk serves as the base ancestor.

## Reference

KeePassXC's three-way merger:
<https://github.com/keepassxreboot/keepassxc/blob/develop/src/core/Merger.cpp>

## Related

- #243
- #153

christhomas:
Closes #337.

## What this adds

`Database::merge_with_ancestor(other, ancestor)` — a three-way merge that uses a common ancestor to distinguish one-sided changes from true conflicts:

| Situation | Resolution |
|-----------|------------|
| Only `self` changed since ancestor | keep `self` silently |
| Only `other` changed since ancestor | take `other` silently |
| Both changed identically | trivially resolved |
| Both changed differently | warn via `MergeLog::warnings`, resolve by newest-wins |

`Database::merge` is now a shim over `merge_with_ancestor(other, &Database::new())`. An empty ancestor makes every object appear changed on both sides, which degrades to the existing two-way semantics with no new warnings — fully backward-compatible.

## Implementation

The ancestor is threaded through the existing `merge_groups` / `merge_entries` helpers as a third `&Database` parameter. In the "both exist" case for entries and groups, both sides are compared against the ancestor's `last_modification` timestamp before falling through to the existing newest-wins logic.

## Tests

Five new tests alongside the existing 31:

- `test_ancestor_one_sided_change_in_other` — no warning, other's value taken
- `test_ancestor_one_sided_change_in_self` — no warning, self's value kept
- `test_ancestor_true_conflict_warns_and_newest_wins` — one warning, newest wins, history preserved
- `test_ancestor_idempotence`
- `test_two_way_shim_no_warning_on_conflict` — backward-compat: `merge()` produces no new warnings

Run with `cargo test --lib --features _merge`.

### DIFF
--- a/src/db/merge.rs
+++ b/src/db/merge.rs
@@ -2612,4 +2612,106 @@ mod merge_tests {
             Some("entry1_updated_in_other"),
         );
     }
+
+    /// Test that when both sides move an entry to different groups, the side with the more
+    /// recent modification wins both the location and the content.
+    ///
+    /// Scenario:
+    ///   ancestor: entry1 in root
+    ///   self:     moves entry1 → group1, updates content  (earlier)
+    ///   other:    moves entry1 → group2, updates content  (later)
+    /// Expected: entry1 ends up in group2 with other's content; one conflict warning.
+    #[test]
+    fn test_ancestor_entry_moved_to_different_groups_other_wins() {
+        let ancestor_db = create_test_database();
+        let mut self_db = ancestor_db.clone();
+        let mut other_db = ancestor_db.clone();
+
+        sleep();
+
+        // self moves entry1 to group1 and updates content
+        self_db
+            .entry_mut(ENTRY1_ID)
+            .unwrap()
+            .track_changes()
+            .move_to(GROUP1_ID)
+            .expect("move to group1");
+        self_db.entry_mut(ENTRY1_ID).unwrap().edit_tracking(|e| {
+            e.set_unprotected(fields::TITLE, "entry1_moved_to_group1_by_self");
+        });
+
+        sleep();
+
+        // other moves entry1 to group2 and updates content (newer)
+        other_db
+            .entry_mut(ENTRY1_ID)
+            .unwrap()
+            .track_changes()
+            .move_to(GROUP2_ID)
+            .expect("move to group2");
+        other_db.entry_mut(ENTRY1_ID).unwrap().edit_tracking(|e| {
+            e.set_unprotected(fields::TITLE, "entry1_moved_to_group2_by_other");
+        });
+
+        let result = self_db.merge_with_ancestor(&other_db, &ancestor_db).unwrap();
+
+        // true conflict: both sides changed since ancestor — our warning must be present
+        assert!(
+            result
+                .warnings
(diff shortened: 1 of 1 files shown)


christhomas:
I don't have any files which have conflicts, so can somebody try this and see whether it gives the right output, or instruct me on how I could create a test case for it and I'll see how to implement it.

christhomas:
The CI failure is unrelated to this PR — the 'Test Suite on x86_64' job failed at the 'Install cross-tool test dependencies' step because `curl` timed out trying to reach keepass.info (`curl: (28) Failed to connect to keepass.info port 443 after 134184 ms`). Rerunning should clear it.

droidmonkey:
Can you add a test for entries that are in different groups between Source and Destination. Ideally the last modified entry would win the location battle.

christhomas:
@droidmonkey  there you go

droidmonkey (dismissed):

sseemayer:
Thanks for the contribution. I am surprised that the changeset is this small. Does this really cover all of the cases of three-way merging? Have you spent the time thinking how the merge logic should work, or is this just a direct implementation of #337 ?

I want to spend some time, probably this weekend to properly think through the merge logic from first principles.

christhomas:
you give me a clear direction on the gap you think that we have and I'll go away and fix it. I have to admit I'm not an expert at keepassxc, so I have no choice but to delegate some level of direction to you. But I like the project and I use it everyday, so I'm using my free time for good use :)

christhomas:
I added some extra situations which might be useful, do you want to look into it please?

PR #344, opened 2026-05-29
Expose merge types

### LINKED ISSUE #343: Feat: expose `merge` types
Request: re-export `mod merge` as `pub mod merge` (MergeLog / MergeError unreachable with `_merge` feature on).

With the `_merge` feature enabled, the merge implementation compiles but its types (`MergeLog`, `MergeError`, `MergeEvent`, `MergeEventType`, `MergeEventTarget`) are not reachable from outside the crate.

e.g.

`Database::merge`:

```rust
pub fn merge(&mut self, other: &Database) -> Result<MergeLog, MergeError>
```

returns types whose names downstream crates cannot bring into scope. Calling `.merge(...)` from outside is possible only by binding the return value to `_` and ignoring the merge log.

## Suggested fix

```rust
#[cfg(feature = "_merge")]
pub mod merge;
```

No semantic change; just opens the existing types to crates that opt into the feature.

Will open a PR with the proposed implementation momentarily.

## Related

#280 (same re-export pattern requested for `VariantDictionaryValue`)

davidroeca:
Re-export the merge module so its result types are nameable downstream.

Happy to remove the test commit if it's overkill. Just say the word.

`Database::merge` is `pub`, but the `merge` module is private, so `MergeLog`, `MergeError`, `MergeEvent`, `MergeEventType`, and `MergeEventTarget` cannot be brought into scope outside the crate even with the `_merge` feature enabled. Callers can currently only bind the return value to `_`.

This mirrors the existing `pub mod fields;` and is gated behind the `_merge` feature, so default builds are unaffected.

~Waiting on #342 before proceeding with doc updates to pass CI~

Closes: https://github.com/sseemayer/keepass-rs/issues/343


louib (commented):
I honestly think the test is overkill. What we could add instead is an example so that people are aware of the merge feature even if it is hidden. What do you think?

davidroeca:
Yeah that sounds reasonable to me. Where would we want that example to go?

louib:
@davidroeca 

[right here with the other examples](https://github.com/sseemayer/keepass-rs/tree/master/examples)

examples/merge_databases.rs
```diff
@@ -0,0 +1,86 @@
...
+        println!("Merge completed with {} warning(s):", log.warnings.len());
+        for warning in &log.warnings {
+            println!("  - {warning}");
+        }
+    }
+
+    // Categorise each change by matching on the public event enums.
+    println!("Applied {} change(s):", log.events.len());
+    for MergeEvent { target, event_type } in &log.events {
+        let kind = match event_type {
```
louib, line 59: @davidroeca (not a blocker) this could be made into a function of `MergeEventType`

examples/merge_databases.rs
```diff
@@ -0,0 +1,86 @@
...
+            _ => "changed",
+        };
+        let what = match target {
+            MergeEventTarget::Entry(id) => format!("entry {id}"),
+            MergeEventTarget::Group(id) => format!("group {id}"),
+            MergeEventTarget::Icon(id) => format!("icon {id}"),
+            // `MergeEventTarget` is `#[non_exhaustive]`.
+            _ => "object".to_string(),
+        };
+        println!("  - {kind} {what}");
```
louib, line 74: @davidroeca (not a blocker) the whole pretty-print logic could be made into a function of `MergeEvent`

louib (approved):

PR #348, opened 2026-06-07
ci: enable github-native code coverage

louib:
I'm not sure it is feature-rich enough to replace codecov just yet but I think it's worth a try
I followed the documentation [here](https://docs.github.com/en/code-security/how-tos/maintain-quality-code/set-up-code-coverage)


varjolintu (approved):

louib:
Closing since we cannot enable Code Quality on non-org repos

PR #358, opened 2026-07-05
fix: correct several cross-library compatibility issues

### LINKED ISSUE #357: Incompatibility with other libraries after Database.save() caused by AES-KDF UUID
The `KDF_AES_KDBX4` UUID from [config.rs:392](https://github.com/sseemayer/keepass-rs/blob/f3262d4b697d11cda1a48982b871b3c980087e8e/src/config.rs#L392) isn't really mentioned in KDBX [spec](https://keepass.info/help/kb/kdbx.html).

That may led to inability to read the database by other libraries, as UUID won't be recognized. I'm not sure which libraries will become incompatible, but [kotpass](https://github.com/keemobile/kotpass) definitely incompatible, [pykepass](https://github.com/libkeepass/pykeepass) - not sure, but it doesn't have this UUID.

It looks like, this UUID was introduced in this KeePassXC [PR](https://github.com/keepassxreboot/keepassxc/pull/1230). But it seems to be not used, as there is a comment in [AesKdf.cpp:55](https://github.com/keepassxreboot/keepassxc/blob/ce28e2afb7a59ab5c4b64512b7119eec4a9f2008/src/crypto/kdf/AesKdf.cpp#L55): `always write old KDBX3 AES-KDF UUID for compatibility with other applications`.

Considering all above, it could be a good idea to write old `KDF_AES_KDBX3` UUID and use `KDF_AES_KDBX4` only for reading.

sseemayer:
closes #357 

- this crate was using a non-spec-compliant UUID to denote the AES Key Derivation function. Now use the standard UUID of C9D9F39A628A4460BF740D08C18A4FEA when dumping databases
- add missing support for legacy hex keyfiles
- extend the cross-tool tests to perform matrix testing on different DatabaseConfig combinations, ensuring the resulting databases can be parsed by keepass and keepassxc

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -352,7 +352,8 @@ impl KdfConfig {
 
         match self {
             KdfConfig::Aes { rounds } => {
-                vd.set(KDF_ID, KDF_AES_KDBX4.to_vec());
+                // always use the KDBX3 AES KDF UUID for compatibility with other libraries
+                vd.set(KDF_ID, KDF_AES_KDBX3.to_vec());
                 vd.set(KDF_ROUNDS, *rounds);
                 vd.set(KDF_SEED, seed.to_vec());
             }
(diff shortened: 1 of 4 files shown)


sseemayer:
The CI/CD failure is unrelated to the code changes and will be fixed by #359

louib (approved):

droidmonkey (approved):

droidmonkey:
Might be worth looking at the keepassxc tests for key files, they are really annoying.

PR #362, opened 2026-07-13
feat: make ID constructors public

louib:
This allows creating new IDs without having to import the `uuid` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ chrono = { version = "0.4.23", default-features = false, features = [
 rust-argon2 = "3.0"
 sha1 = { version = "0.11", optional = true }
 sha2 = "0.11"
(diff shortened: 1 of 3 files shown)


sseemayer (dismissed):

Cargo.toml
```diff
@@ -54,7 +54,7 @@ chrono = { version = "0.4.23", default-features = false, features = [
 rust-argon2 = "3.0"
 sha1 = { version = "0.11", optional = true }
 sha2 = "0.11"
-aes = "0.9"
+aes = "<0.9.3"
```
louib, line 57: Looks like aes 0.9.3 broke compatibility with rustc `1.85.1`. This will be a breaking change for our crate so let's defer that to a separate PR.

sseemayer (approved):

PR #363, opened 2026-07-14
feat: preserve group and entry insertion order in group

aivanovski:
**Description of issue**
The order of groups and entries written in XML is determined by function `db_to_xml`, that iterates over `Group.groups` and `Groups.entries`. The previous implementation kept child group and entry IDs in `HashSet`s, so iteration order was not guaranteed. As a result, save/open round trips could reorder groups or entries even when no user-visible ordering change was intended.

**Description of changes**
`HashSet` were replaced with `IndexSet`.

Tests were added to check:
- root entry order survives saving and reopening
- order is still preserved after deleting selected groups and entries

_As I’m still new to Rust, so I’d really appreciate any feedback or suggestions._

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ cbc = "0.2"
 challenge_response = { version = "0.5", optional = true, default-features = false, features = ["nusb"] }
 
 uuid = { version = "1.18.1", features = ["v4", "serde"] }
+indexmap = { version = "2", features = ["serde"] }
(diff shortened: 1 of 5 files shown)


sseemayer:
Thanks for your contribution! It looks good on first glance but I still want to understand the indexmap crate better, so I will look at this in more detail on the weekend hopefully.

sseemayer (approved):

PR #367, opened 2026-08-15
Fix DeletedObject DeletionTime XML serialization

dmshvedchenko:
## Summary

Fix `DeletedObject` XML handling so KDBX saves use the schema-defined `DeletionTime` element, while still accepting files previously written with the incorrect Rust field name `deletion_time`.

The change also prevents the writer from emitting invalid tombstones when the deletion timestamp is already missing. Valid tombstones are preserved unchanged.

## Root cause

`DeletedObject::uuid` explicitly maps to the KDBX `UUID` element, but `deletion_time` had no Serde rename:

```rust
#[serde(default, with = "cs_opt_string")]
deletion_time: Option<Timestamp>,
```

This caused two related failures:

1. Canonical `<DeletionTime>` values were not mapped to the field and defaulted to `None`.
2. Saving then emitted the Rust field name as `<deletion_time/>` instead of the schema-defined `<DeletionTime>`.

Applications with strict KDBX XML validation, including KeePassium, reject the resulting database.

## Changes

- Serialize `deletion_time` as `DeletionTime`.
- Accept legacy `deletion_time` during deserialization so databases written by affected clients can still be opened and repaired.
- Omit `DeletedObject` values whose timestamp is `None` instead of emitting an invalid empty `DeletionTime` element.
- Preserve every tombstone that still has a valid timestamp.

## Damaged-file behavior

If an affected database was already saved by the old writer, the original deletion timestamps may have been replaced by empty `<deletion_time/>` elements. Those timestamps cannot be reconstructed from that copy.

For such entries, omitting the invalid tombstone is preferable to generating XML that other KeePass clients cannot open. This only removes already-incomplete deletion-sync metadata; it does not remove password entries, groups, or entry history. Callers that repair an existing file should still write to a separate output file and retain the original backup.

## Tests

Added regression coverage for:

- direct serialization emits `<DeletionTime>` and never `<deletion_time>`;
- legacy non-empty `<deletion_time>` is accepted;
- legacy empty `<deletion_time/>` is accepted as a missing timestamp;
- an encrypted KDBX save contains the canonical tag;
- tombstones without timestamps are omitted and the saved KDBX reopens successfully;
- existing valid `DeletedObjects` round-trip coverage continues to pass.

Validation performed:

```text
cargo fmt --all
cargo test --features "save_kdbx4 totp"
cargo build --features "save_kdbx4 totp"
```

The full test suite passed. A consuming Jaiba build also passed its tests and release build.

## Manual interoperability validation

The complete affected-file workflow was also validated manually:

1. A damaged KDBX copy containing empty deletion tombstones was repaired to a separate output file.
2. The repaired copy opened successfully in both a consuming application and a strict external KeePass client.
3. The database was modified and saved again by the consuming application.
4. The strict external client successfully reopened the newly saved file.

This revisits the issue reported in #302, adding backward-compatible legacy parsing, damaged-file handling, and regression coverage.

### DIFF
--- a/src/format/xml_db/mod.rs
+++ b/src/format/xml_db/mod.rs
@@ -293,7 +293,12 @@ pub struct DeletedObject {
     #[serde(rename = "UUID")]
     uuid: UUID,
 
-    #[serde(default, with = "cs_opt_string")]
+    #[serde(
+        default,
+        rename = "DeletionTime",
+        alias = "deletion_time",
+        with = "cs_opt_string"
+    )]
     deletion_time: Option<Timestamp>,
 }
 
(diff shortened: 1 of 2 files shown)


sseemayer (approved):

sseemayer:
thanks for spotting this and the fix, and welcome! 🚀

PR #368, opened 2026-08-20
fix(DeletionTime): unify timestamp serialization mode

GodCount:
Unify timestamp serialization by using the From trait.

### DIFF
--- a/src/format/xml_db/mod.rs
+++ b/src/format/xml_db/mod.rs
@@ -226,7 +226,7 @@ impl KeePassFile {
                     .iter()
                     .map(|(uuid, deletion_time)| DeletedObject {
                         uuid: UUID(*uuid),
-                        deletion_time: deletion_time.map(Timestamp::new_iso8601),
+                        deletion_time: deletion_time.map(Timestamp::from),
                     })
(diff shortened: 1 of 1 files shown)


sseemayer:
@GodCount  Thanks for the PR! Can you give a bit more background on the "why"? Is it fixing a specific database parsing issue? Could a test be added to prevent a future regression?

GodCount:
> Can you give a bit more background on the "why"?

It is mentioned here.
https://github.com/sseemayer/keepass-rs/blob/2f1dd5e0f1a23dc7420c3fa25f434fe362729b24/src/format/xml_db/timestamp.rs#L46-L51

>  Is it fixing a specific database parsing issue?

No! It's just for format consistency. I also checked other KeePass clients, such as [KeePassXC](https://github.com/keepassxreboot/keepassxc), and they are also in base64 format.

> Could a test be added to prevent a future regression?

Added. If the test doesn't meet the requirements, feel free to @ me.

sseemayer (approved):

sseemayer:
alright, makes sense! Thanks for your contribution!

